PR #116, opened 2026-03-27
fix(wsl2): Fix all errors in wsl env

AlejandroRomanIbanez:
# Fix: Mailu Admin Gunicorn Startup & API Token 502 Errors (DinD/WSL2)

Fixes the issue where the Mailu admin gunicorn process never starts in Docker-in-Docker (DinD) or WSL2 environments due to DNSSEC validation failures in the unbound resolver. Also resolves persistent 502 errors when fetching API tokens caused by circular proxy routing.

---

### 🛠 Template Type
- **Fix** - Repairs broken or incorrect server behavior
- **Feature** - Adds or extends server functionality (via environment-specific DNS forwarding)

---

### 🏗 Affected Roles and Services
- **Primary web role:** `web-app-mailu`
- **Related role:** `resolver` (unbound) container within the Mailu compose stack

---

### 📈 Change Type
- **Patch** - Small improvement or compatible adjustment

---

#### **The Problem**
In **DinD/WSL2** environments, Mailu's admin container `start.py` runs `test_DNS()` before launching gunicorn. This function loops indefinitely waiting for the Mailu unbound resolver to return DNSSEC-validated responses (AD flag). 

In DinD, unbound attempts iterative resolution from root servers, which typically fails because outbound DNS networking is unreliable or restricted. Consequently:
1. **Gunicorn never starts** on port `:8080`.
2. All API token operations fail.
3. A secondary issue: `MAILU_API_SERVICE_BASE_URL` was routing via the `front` nginx, creating a **circular proxy route** that returns 502s because the admin backend is unresponsive.

#### **Root Cause Chain**
1. **DinD degrades iterative DNS** → Unbound DNSSEC validation fails → `test_DNS()` loops → Gunicorn never starts.
2. `http://front/api/v1` → Front Nginx → **502 Bad Gateway** (Admin backend is down).
3. Without `--fail`, `curl` exits with code 0 on HTTP errors → The 502 HTML passes the `until` guard → `from_json` crashes.

#### **Detailed Changes**

| File | Change |
| :--- | :--- |
| `templates/unbound.conf.j2` | Adds `forward-zone` to `1.1.1.1`/`8.8.8.8` when `DOCKER_IN_CONTAINER` is true, enabling DNSSEC via forwarders. |
| `templates/compose.yml.j2` | Mounts the rendered `unbound.conf` over `/unbound.conf` in the resolver container. |
| `tasks/01_core.yml` | Creates override directory and renders `unbound.conf.j2` before deployment. |
| `vars/main.yml` | Updates `MAILU_API_SERVICE_BASE_URL` to `http://localhost:8080/api/v1` to avoid circular proxying. |
| `tasks/03a_manage_user_token.yml` | Added `--fail` to `curl` and `| trim` to the `until` guard to ensure robust retries on failure. |

---

### 🔄 Alternatives Considered
* **Changing admin DNS to 1.1.1.1:** Rejected. Docker prepends `127.0.0.11` in `/etc/resolv.conf`, so `test_DNS()` still hits the failing local resolver.
* **Pre-rendered unbound.conf:** Rejected. This would bypass unbound’s internal Jinja2 rendering of critical `SUBNET` variables.
* **Increasing retry count:** Rejected. The issue is structural (process never starts), not a timing race.

---

### 🌍 Impact & Security
* **Production (non-DinD):** Zero impact. The `forward-zone` is only injected if `DOCKER_IN_CONTAINER | bool` is **True**.
* **Performance:** `localhost:8080` is more direct than routing through the Nginx `front` container.
* **Security:** No negative impact. DNSSEC validation is still performed by unbound; the forwarder acts as an upstream resolver, not a trusted bypass.

---

### ✅ Local Validation
- **Environment:** WSL2 / Docker-in-Docker
- **Result:** `make test-local-app` passed.
- **Verification:** Confirmed Mailu admin starts, blackhole users are created, and tokens are provisioned successfully.

---

### 🔍 Review Focus
- **Jinja2 Escaping:** `templates/unbound.conf.j2` uses `{% raw %}` blocks to ensure Unbound's native Jinja syntax isn't consumed by Ansible.
- **Port Mapping:** Gunicorn port `8080` was verified by inspecting `ghcr.io/mailu/admin:2024.06`.

---

### 💡 Additional Notes
The gunicorn port is currently hardcoded based on the 2024.06 image. If the Mailu version is updated in the future, we should consider extracting this port into a global variable.

### DIFF
--- a/roles/web-app-mailu/templates/compose.yml.j2
+++ b/roles/web-app-mailu/templates/compose.yml.j2
@@ -6,6 +6,8 @@
     image: {{ MAILU_DOCKER_FLAVOR }}/unbound:{{ MAILU_VERSION }}
     container_name: {{ MAILU_CONTAINER }}_resolver
 {% include 'roles/sys-svc-container/templates/base.yml.j2' %}
+    volumes:
+      - "{{ MAILU_UNBOUND_CONF_FILE }}:/unbound.conf:ro"
 {% include 'roles/sys-svc-container/templates/networks.yml.j2' %}
         ipv4_address: {{ MAILU_DNS_RESOLVER }}
 
--- a/roles/web-app-mailu/tasks/01_core.yml
+++ b/roles/web-app-mailu/tasks/01_core.yml
@@ -3,6 +3,18 @@
 - name: "Disable SMTP to avoid port conflicts"
   include_tasks: "{{ [playbook_dir, 'roles/sys-svc-mail-smtp/tasks/disable.yml' ] | path_join }}"
 
+- name: Ensure Mailu unbound overrides directory exists
+  file:
+    path: "{{ MAILU_UNBOUND_CONF_FILE | dirname }}"
+    state: directory
+    mode: "0755"
+
+- name: Render Mailu unbound config override
+  template:
+    src: unbound.conf.j2
+    dest: "{{ MAILU_UNBOUND_CONF_FILE }}"
+    mode: "0644"
+
 - name: Ensure Rspamd overrides directory exists (host)
   file:
     path: "{{ MAILU_RSPAMD_HOST_DIR }}"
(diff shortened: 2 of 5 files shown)


roles/web-app-mailu/templates/compose.yml.j2
kevinveenbirkenbach: Seems all good under the premise that the unbound file is correct 👍

PR #189, opened 2026-04-22
fix(mattermost): inject SSO button via JS for v11 Team Edition

AlejandroRomanIbanez:
## Summary

- Mattermost v11 Team Edition does not render the GitLab SSO button natively: `EnableSignInWithGitLab` is absent from the client config API even when `MM_GITLABSETTINGS_ENABLE=true`, so the React frontend never renders the button
- Extends `javascript.js.j2` to inject an `<a href="/oauth/gitlab/login">SSO with Infinito.Nexus</a>` button via `MutationObserver` when the login form is detected (`#input_loginId` present) but no native SSO anchor exists
- Button uses Mattermost's own CSS classes (`btn btn-primary login-body-card-form-button-submit large`) plus explicit flex/width/margin styles to match the native "Log in" button alignment exactly
- Updates `playwright.spec.js`: `startMattermostSsoFlow` now navigates to `/login`, handles the v11 `/landing` redirect for fresh browser contexts, then clicks the injected button — proving both the JS injection and the OAuth2 backend in a single flow

## Test plan

- [x] Deploy `web-app-mattermost` and hard-refresh `/login` — "SSO with Infinito.Nexus" button appears below "Log in", aligned and spaced correctly
- [x] Click the button — redirects to Keycloak, SSO login completes, lands on Mattermost channel view
- [x] Email/password login still works (untouched)
- [x] Playwright E2E passes: `make test-e2e-playwright -e TEST_E2E_PLAYWRIGHT_ONLY_ROLES='["web-app-mattermost"]'`

### DIFF
--- a/roles/web-app-mattermost/files/playwright.spec.js
+++ b/roles/web-app-mattermost/files/playwright.spec.js
@@ -60,12 +60,22 @@ async function performOidcLogin(frame, username, password) {
   await signInButton.click();
 }
 
-// Navigate directly to the Mattermost GitLab OAuth2 login endpoint.
-// This bypasses the login page button rendering (which requires EnableSignInWithGitLab
-// in the client config) and also bypasses the /landing app-selection dialog that
-// Mattermost shows for fresh browser contexts in v11+.
+// Navigate to the Mattermost login page and click the SSO button injected by javascript.js.j2.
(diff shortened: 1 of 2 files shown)


kevinveenbirkenbach (approved):
@AlejandroRomanIbanez trusting you that it's tested and works. Code looks good.

PR #225, opened 2026-05-01
fix(web-app-pihole): fix compose template, RBAC group, playwright env and wait flag

pragepani:
Removed duplicate networks: block from compose template
Fixed redis volume declaration
Switched to TCP healthcheck
Fixed users.x.y → lookup('users', 'x').y in playwright.env.j2
Removed manual network creation from 01_core.yml
Added TEST_E2E_PLAYWRIGHT_WAIT_ENABLED: false in vars
Fixed RBAC group name to match LDAP (web-app-pihole-administrator)
Fixed allowed_groups to use platform convention with RBAC.GROUP.NAME


pragepani:
Screenshot 2026-05-01 at 11 19 11 AM
Screenshot 2026-05-01 at 11 20 03 AM

PR #231, opened 2026-05-05
feat(svc-runners): provision self-hosted CI runners with cache, isolation, and timeout scaling

AlejandroRomanIbanez:
## Summary

Adds `svc-runner` to provision a machine as parallel GitHub Actions self-hosted CI runner instances, and hardens the CI deploy scripts to run reliably on self-hosted hardware.

---

## Template Type

* [x] **Feature** - Adds or extends shared system functionality
* [x] **Fix** - Repairs broken or incorrect shared system behavior

---

## Affected Components

* Primary role(s) or module(s): `svc-runner`
* Related services, timers, packages, inventories, or workflows: `scripts/tests/deploy/ci/dedicated.sh`, `scripts/tests/deploy/ci/multiply-timeouts.sh`, `scripts/meta/env/defaults.sh`, `compose.yml`, `Makefile`, `test-deploy-server.yml`, `test-deploy-universal.yml`, `test-deploy-workstation.yml`
* Distro(s) tested: Debian

---

Optional. Ignored unless a maintainer applies the **🧩 Subset** label; without the label CI uses the diff-derived role set as usual. When the label is set, CI deploys **only** the roles listed here — each must be an existing `roles/<id>` directory, or the run fails. This gives fork PRs the manual-dispatch whitelist. See [pipeline.md](../../docs/contributing/artefact/git/pipeline.md#subset-label-).

```yaml
roles:
  - svc-runner
```

---

## Change Type

* [x] **Patch** - Small improvement or compatible adjustment

---

## Change Details

**`svc-runner` role**

Provisions N isolated runner instances on the same host. Each instance gets a dedicated Docker volume (`INFINITO_DOCKER_VOLUME`), network subnet, DNS IP, and container name prefix (`INFINITO_RUNNER_PREFIX`) so parallel jobs cannot collide with each other or with any sibling process on the host.

**Image cache vs. credential freshness**

Self-hosted runners set `INFINITO_PRESERVE_DOCKER_CACHE=true` so inner-Docker image layers survive between jobs — avoiding repeated pulls of large images. Only `volumes/` and `containers/` are wiped before each run so database credentials are always fresh. The wipe runs before `compose up` to avoid a race with the inner Docker daemon.

**Timeout scaling**

Self-hosted hardware may have slower storage than GitHub-hosted runners. `multiply-timeouts.sh` scales Ansible `retries:` and Docker `start_period:` values inside the ephemeral container by `INFINITO_TIMEOUT_MULTIPLIER`. The value is driven by the repo variable `vars.INFINITO_TIMEOUT_MULTIPLIER` (set via `make runner-ci-enable`) — no values are hardcoded in workflow files. The fallback is `1` (no-op), so GitHub-hosted jobs are unaffected.

**`STORAGE_CONSTRAINED` guard**

The inner Docker daemon reports its root on a small tmpfs mount. The CLI storage auto-detect would treat this as a constrained environment and prune images, defeating the cache. `--force-storage-constrained false` is passed to `cli.deploy.development init` when `INFINITO_PRESERVE_DOCKER_CACHE=true`.

**`INFINITO_RUNNER_PREFIX` in `defaults.sh`**

Integrated into the existing `INFINITO_CONTAINER` SPOT so the single-source-of-truth guarantee is preserved. Default is `infinito` — identical behaviour to before on any setup that does not set the variable.

---

## Local Validation

* [x] `svc-runner` deployed — 2 runner instances registered and active
* [x] Full deploy matrix passed across all distros with `INFINITO_TIMEOUT_MULTIPLIER=30`
* [x] GitHub-hosted runner jobs confirmed unaffected

---

## E2E Proof — `svc-runner` DinD Full Deploy (`test-e2e-cli` → `local.sh`)

**Workflow run:**  
https://github.com/AlejandroRomanIbanez/infinito-nexus/actions/runs/27868452284/job/82515564524

> The `svc-runner` job completed successfully. The overall workflow is marked as failed only because of an unrelated application in the matrix.

### What was validated

The runner brings its instances up locally, then deploys web-app-dashboard (disable=matomo, RUNTIME=github) inside runner-1 via DooD and runs the Playwright E2E — i.e. a real-life full-environment deploy on localhost.

### Execution Summary

```text
OK: runner-1 is running
OK: runner-2 is running
OK: runner-4 is running

DinD mode: verifying DooD socket in runner-1...
OK: DooD socket accessible inside runner-1

DinD mode: running full deploy test inside runner-1...

=== Running dedicated distro deploy: distro=debian app=web-app-dashboard ===

...

PLAY RECAP ***************************************************************
localhost : ok=997 changed=... unreachable=0 failed=0 ...

...

Playwright Run Summary:
1 passed (4.6s)

OK: full deploy inside runner-1 succeeded

ALL LOCAL CHECKS PASSED (DinD)
ALL CHECKS PASSED
```

### Result

- ✅ Local runner infrastructure started successfully.
- ✅ DooD socket verified inside `runner-1`.
- ✅ Full deployment of `web-app-dashboard` completed successfully.
- ✅ Playwright E2E test passed.
- ✅ Every Ansible play recap reported `failed=0`.
- ✅ No deployment failures occurred anywhere in the `svc-runner` job.
- ✅ Full localhost deployment validation completed successfully in DinD mode.

## Security Impact

* [x] Security impact present

Runner `.env` files are written with mode `0600`. GitHub registration tokens are fetched at deploy time via `gh api` and never persisted. The volume wipe is scoped to `volumes/` and `containers/` subdirectories only.

---

## Review Focus

* `scripts/tests/deploy/ci/dedicated.sh` — volume wipe order, `--force-storage-constrained false` conditional, label-scoped container cleanup
* `scripts/meta/env/defaults.sh` — `INFINITO_RUNNER_PREFIX` in the `INFINITO_CONTAINER` SPOT
* `roles/svc-runner/tasks/instance.yml` — per-instance `.env` generation and runner registration

---

## Definition of Done (DoD)


compose/Corefile
kevinveenbirkenbach, line 1: This file shouldn't exist. It's autogenerated.

roles/svc-runner/README.md
```diff
@@ -0,0 +1,149 @@
...
+`runner_distribution` is **required** and has no default; it is passed automatically by the CLI.
+
+## Further Resources
+
+- [GitHub Actions self-hosted runner documentation](https://docs.github.com/en/actions/hosting-your-own-runners)
+- [actions/runner releases](https://github.com/actions/runner/releases)
+
+## Credits
+
+Developed and maintained by **Kevin Veen-Birkenbach**.
```
kevinveenbirkenbach, line 156: Your name and your homepage you can add here

roles/svc-runner/meta/main.yml
```diff
@@ -0,0 +1,29 @@
+galaxy_info:
+  author: Kevin Veen-Birkenbach
```
kevinveenbirkenbach: Your name.

kevinveenbirkenbach (commented):
Looks overall good. Please add an github workflow test which reproduces this behaviour if possible.

kevinveenbirkenbach:
Looks good. Just set up an test to reproduce this behaviour, to guaranty that it doesn't break in the future. may an option is to generate under roles/*/files/test.sh an test routine to execute cli commands and tests, equivalent to playwright and the workflows catch them. Not 100% sure what is the best approach.

In this  case the normal test-deploy-server.yml could execute the runner against the deployed runner.

cli/deploy/development/down.py
```diff
@@ -15,7 +15,6 @@ def _repo_root_from_here() -> Path:
 
 
 CI_DOCKER_ROOT = Path("/mnt/docker")
```
kevinveenbirkenbach: Check this out.

kevinveenbirkenbach: Keep SPOT in mind.

cli/deploy/development/down.py
```diff
@@ -40,19 +39,27 @@ def _compose_run(*, repo_root: Path, distro: str, args: list[str]) -> None:
...
+        print(f">>> Not on GitHub - No bind volumes will be deleted: {CI_DOCKER_ROOT}")
+        return
+
+    if os.environ.get("INFINITO_PRESERVE_DOCKER_CACHE", "false").lower() == "true":
+        print(
+            ">>> INFINITO_PRESERVE_DOCKER_CACHE=true — keeping Docker root for next distro"
+        )
         return
 
     docker_root_env = os.environ.get("INFINITO_DOCKER_VOLUME", "").strip().rstrip("/")
```
kevinveenbirkenbach: function with logik,, too stay to SRP, KISS

roles/svc-prx-openresty/tasks/01_core.yml
```diff
@@ -4,6 +4,17 @@
     state: directory
     mode: "0755"
 
+- name: Stat nginx.conf path
```
kevinveenbirkenbach: This seems like it doesn't make sense; Why delete and then copy it ? Copying should be sufficient.

roles/svc-runner/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
+runner_github_owner: kevinveenbirkenbach
```
kevinveenbirkenbach:
```suggestion
runner_github_owner: infinito-nexus
```

roles/svc-runner/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
...
+runner_github_repo: infinito-nexus
```
kevinveenbirkenbach:
```suggestion
runner_github_repo: core
```

roles/svc-runner/defaults/main.yml
kevinveenbirkenbach, line 1: move the configuration to meta/services.yml and define the variables with lookup(config in vars/main.yml.

Because they are constants write them UPPERCASE.

roles/svc-runner/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
...
+runner_name: "{{ inventory_hostname }}"
+runner_distribution: ""
+runner_labels: "self-hosted,linux,{{ runner_distribution }}"
+runner_install_dir: /opt/github-runner
+runner_user: github-runner
+runner_count: 1
```
kevinveenbirkenbach: Auto-Default based on server values.

roles/svc-runner/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
...
+# CPU limit for each runner's infinito container (matches GitHub-hosted 2-core quota).
+# Set to total_host_cpus / runner_count. Prevents one runner from starving others.
+runner_cpus: 2
```
kevinveenbirkenbach: Auto-Default based on server values.

roles/svc-runner/files/test.sh
```diff
@@ -0,0 +1,82 @@
...
+set -euo pipefail
+
+PASS=0
+FAIL=0
+
+ok()   { echo "PASS: $*"; PASS=$((PASS + 1)); }
+fail() { echo "FAIL: $*"; FAIL=$((FAIL + 1)); }
+
+RUNNER_COUNT="${RUNNER_COUNT:-1}"
+RUNNER_INSTALL_DIR="${RUNNER_INSTALL_DIR:-/opt/github-runner}"
```
kevinveenbirkenbach: USE one SPOT no default values.

roles/svc-runner/files/test.sh
```diff
@@ -0,0 +1,82 @@
...
+# Run after `scripts/tests/runner.sh` completes the Ansible deploy.
+set -euo pipefail
+
+PASS=0
+FAIL=0
+
+ok()   { echo "PASS: $*"; PASS=$((PASS + 1)); }
+fail() { echo "FAIL: $*"; FAIL=$((FAIL + 1)); }
+
+RUNNER_COUNT="${RUNNER_COUNT:-1}"
```
kevinveenbirkenbach: USE one SPOT no default values.

roles/svc-runner/meta/services.yml
```diff
@@ -0,0 +1,2 @@
+svc-runner:
+  lifecycle: alpha
```
kevinveenbirkenbach: change it to beta when auto tests pass.

roles/svc-runner/tasks/install/archlinux.yml
kevinveenbirkenbach, line 1: Fedora CentOS missing

roles/svc-runner/tasks/install/archlinux.yml
```diff
@@ -0,0 +1,11 @@
+- name: Install GitHub Actions runner dependencies (Arch Linux)
+  pacman:
```
kevinveenbirkenbach: Use general function, one file , debian ubuntu routines when DEBIAND or UBUNTU, and list per Distro for package isntall task.

roles/svc-runner/tasks/instance.yml
```diff
@@ -0,0 +1,86 @@
...
+  args:
+    chdir: "{{ runner_install_dir }}/{{ runner_instance }}"
+  become: true
+  become_user: "{{ runner_user }}"
+  no_log: "{{ MASK_CREDENTIALS_IN_LOGS | default(true) | bool }}"
+
+- name: Write runner environment file for instance {{ runner_instance }}
+  copy:
+    dest: "{{ runner_install_dir }}/{{ runner_instance }}/.env"
+    content: |
```
kevinveenbirkenbach: If possible genarelize it. Think about it. If you change nothing its also ok.

roles/svc-runner/tasks/instance.yml
```diff
@@ -0,0 +1,86 @@
...
+- name: Remove existing runner configuration for instance {{ runner_instance }}
+  command: find {{ runner_install_dir }}/{{ runner_instance }} -maxdepth 1 -name ".*" -type f -delete
+  changed_when: false
+
+- name: Get GitHub Actions runner registration token for instance {{ runner_instance }}
+  command: gh api repos/{{ runner_github_owner }}/{{ runner_github_repo }}/actions/runners/registration-token --method POST -q .token
+  delegate_to: localhost
+  register: runner_registration_token
+  changed_when: true
+  no_log: "{{ MASK_CREDENTIALS_IN_LOGS | default(true) | bool }}"
```
kevinveenbirkenbach: Remove the redundant | default( everywhere. it is defined.

roles/svc-runner/tasks/main.yml
```diff
@@ -0,0 +1,154 @@
...
+      - --filter
+      - "label=com.docker.compose.project=runner-{{ item }}"
+  register: _orphaned_container_probe
+  changed_when: false
+  failed_when: false
+  loop: "{{ range(runner_count | int + 1, runner_count | int + 50) | list }}"
+
+- name: Remove orphaned runner DIND containers (instances beyond runner_count)
+  ansible.builtin.command:
+    argv: "{{ ['docker', 'rm', '-f'] + item.stdout_lines }}"
```
kevinveenbirkenbach: I think here container instead of docker should be used.
Probably dependency in meta/main.yml should be the sys-svc-container.

roles/svc-runner/tasks/main.yml
```diff
@@ -0,0 +1,154 @@
...
+  ansible.builtin.mount:
+    path: "/mnt/docker/{{ item }}"
+    state: unmounted
+  loop: "{{ range(runner_count | int + 1, runner_count | int + 50) | list }}"
+  failed_when: false
+  changed_when: false
+
+- name: Remove orphaned Docker volume directories (instances beyond runner_count)
+  ansible.builtin.file:
+    path: "/mnt/docker/{{ item }}"
```
kevinveenbirkenbach: This hardcoded path is going against SPOT.

scripts/meta/env/defaults.sh
```diff
@@ -14,8 +14,9 @@ set -euo pipefail
 # INFINITO_DISTRO — see the failure on web-app-wordpress @ fedora in CI run
 # 24942172018 / job 73038526513.
 : "${INFINITO_DISTRO:=debian}"
-export INFINITO_DISTRO
-export INFINITO_CONTAINER="infinito_nexus_${INFINITO_DISTRO}"
+: "${INFINITO_RUNNER_PREFIX:=infinito}"
```
kevinveenbirkenbach: no default value

scripts/tests/deploy/ci/custom/svc-runner.sh
kevinveenbirkenbach, line 1: shouldn't this be covered by files/test.sh ?

scripts/tests/deploy/ci/dedicated.sh
```diff
@@ -104,23 +111,32 @@ cleanup() {
...
-	# Important for serial multi-distro CI runs on the same runner.
-	docker image prune -af >/dev/null 2>&1 || true
-	docker buildx prune -af >/dev/null 2>&1 || true
-	docker builder prune -af >/dev/null 2>&1 || true
+	# Important for serial multi-distro CI runs on GitHub runners (limited disk).
+	# Skipped on self-hosted (INFINITO_PRESERVE_DOCKER_CACHE=true) so outer images
+	# (infinito, coredns) stay in the Docker cache.  Combined with pull_policy:always
+	# they become fast manifest checks (~5 s) instead of full 3-5 min GHCR pulls
+	# for every subsequent distro run and for every subsequent job.
+	if [[ "${INFINITO_PRESERVE_DOCKER_CACHE:-false}" != "true" ]]; then
```
kevinveenbirkenbach: define the variable somewhere as SPOT, prefered no default fallbacks.

scripts/tests/deploy/ci/dedicated.sh
```diff
@@ -131,24 +147,48 @@ cleanup() {
...
 	echo ">>> HARD cleanup finished"
 	return $rc
 }
 trap cleanup EXIT
 
+# Wipe stale inner-Docker state before starting so credentials are always fresh.
+# Done before `up` to avoid a race with the inner Docker daemon.
+# Image layers (overlay2/, image/) are preserved for cache; only volumes/ and
+# containers/ are removed so old DB passwords and auto-restart state don't survive.
+if [[ "${INFINITO_PRESERVE_DOCKER_CACHE:-false}" == "true" ]]; then
```
kevinveenbirkenbach: No default fallback. This fallback occures multiple  times. please define SPOT.

scripts/tests/deploy/ci/dedicated.sh
```diff
@@ -160,10 +200,24 @@ df -h || true
...
 
+_init_args=(
+	--apps "${APPS}"
+	--inventory-dir "${INVENTORY_DIR}"
+)
+if [[ "${INFINITO_PRESERVE_DOCKER_CACHE:-false}" == "true" ]]; then
+	_init_args+=(--force-storage-constrained false)
+fi
+
+if [[ "${INFINITO_TIMEOUT_MULTIPLIER:-1}" -gt 1 ]]; then
```
kevinveenbirkenbach: No defauult please use SPOT

scripts/tests/deploy/ci/dedicated.sh
```diff
@@ -160,10 +200,24 @@ df -h || true
 docker system df || true
 echo ">>> END STATE BEFORE DEPLOY"
 
+_init_args=(
+	--apps "${APPS}"
+	--inventory-dir "${INVENTORY_DIR}"
+)
+if [[ "${INFINITO_PRESERVE_DOCKER_CACHE:-false}" == "true" ]]; then
```
kevinveenbirkenbach: No defauult please use SPOT

scripts/tests/deploy/ci/multiply-timeouts.sh
```diff
@@ -0,0 +1,36 @@
...
+#
+# Changes (ephemeral — actions/checkout resets them on the next job):
+#   • Ansible task retries      → retries × multiplier
+#   • Docker healthcheck start_period → start_period × multiplier
+#
+# Example: multiplier=30, retries=60, delay=2s → 60×30×2s = 3600s = 1 h
+# Fast hardware finishes early; slow hardware waits as long as needed up to 1 h.
+set -euo pipefail
+
+MULTIPLIER="${1:-1}"
```
kevinveenbirkenbach: No defauult please use SPOT

scripts/tests/deploy/ci/multiply-timeouts.sh
```diff
@@ -0,0 +1,36 @@
...
+# Resolve repo root via the dedicated deploy CLI (realpath → baked-in container path).
+# Falls back to /opt/src/infinito when run outside the container.
+REPO_ROOT="${2:-$(python3 -c 'from cli.deploy.dedicated.paths import REPO_ROOT; print(REPO_ROOT)' 2>/dev/null || echo '/opt/src/infinito')}"
```
kevinveenbirkenbach: No defauult please use SPOT

scripts/tests/runner.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+# Deploys the svc-runner Ansible role to localhost and runs the verification suite.
+# Requires: gh CLI authenticated (GH_TOKEN env var), ansible-playbook in PATH.
+set -euo pipefail
+
+RUNNER_COUNT="${RUNNER_COUNT:-1}"
```
kevinveenbirkenbach: No defauult please use SPOT

scripts/tests/runner.sh
```diff
@@ -0,0 +1,38 @@
...
+RUNNER_DISTRO="${RUNNER_DISTRO:-ubuntu}"
```
kevinveenbirkenbach: No defauult please use SPOT

kevinveenbirkenbach (commented):
Sorry seems like I forgot to submit my review  from last week.

kevinveenbirkenbach:
image

.github/workflows/test-deploy-server.yml
kevinveenbirkenbach, line 1: Use test-e2e-cli role approach to deploy against the test.sh script against the localhost on which the runner is  setup. In CLI runner should be deployed against localhost and the test should also run in localhost.

.github/workflows/test-deploy-universal.yml
kevinveenbirkenbach, line 1: Use test-e2e-cli role approach to deploy against the test.sh script against the localhost on which the runner is  setup. In CLI runner should be deployed against localhost and the test should also run in localhost.

.github/workflows/test-deploy-workstation.yml
kevinveenbirkenbach, line 1: Use test-e2e-cli role approach to deploy against the test.sh script against the localhost on which the runner is  setup. In CLI runner should be deployed against localhost and the test should also run in localhost.

kevinveenbirkenbach (commented):
Image

This is the main issue. refactor to a test-e2e-cli role. 

This should start all roles/*files/test.sh files by using the parameters from roles/*templates/test.env.j2   to run against localhost. Should be indopent, because the test cyclus has an update and and redeploy part so test will be triggered twice against the same instance. 

Should just run in test environment (CI)

kevinveenbirkenbach:
Please also add docs/agents/action/iteration/runner.md with instructions for the agents how to iterate when runner is used. Also please add a cheat to docs/contributing/tools/agents/cheatsheet.md  how to use the program. @pragepani has a similar issue like you and would profit from it.

kevinveenbirkenbach:
image

https://cloud.infinito.nexus/apps/files/files/265444?dir=/Infinito.Nexus/IT/Architecture/CORE&editing=false&openfile=true

.github/workflows/test-deploy-server.yml
kevinveenbirkenbach, line 1: One spot to test, don't repeat on each web deploy.

kevinveenbirkenbach:
image

Full test routine in test.sh required not in cli runners.

roles/svc-runner/files/test.sh
```diff
@@ -0,0 +1,58 @@
...
+# Load default deploy env (sets TEST_DEPLOY_TYPE, etc.)
+# shellcheck source=scripts/meta/env/all.sh
+source "${REPO_ROOT}/scripts/meta/env/all.sh"
+
+# Unset per-instance networking vars — production-only subnet assignments
+# that conflict with the CI test stack which uses env.ci defaults (172.30.0.x).
+unset SUBNET GATEWAY DNS_IP IP4 BIND_IP COMPOSE_PROJECT_NAME INFINITO_RUNNER_PREFIX
+
+# Deploy a real app through the runner's Docker environment to prove it works end-to-end
+APPS=web-app-matomo make -C "${REPO_ROOT}" deploy-fresh-purged-apps
```
kevinveenbirkenbach: It seems like you don'Ät execute the runner. Test the env variables bnefore you execute  to make sure that you actually use the runner.

roles/svc-runner/tasks/instance.yml
```diff
@@ -0,0 +1,97 @@
...
+  no_log: "{{ MASK_CREDENTIALS_IN_LOGS | bool }}"
+  when: not DOCKER_IN_CONTAINER
+
+- name: Write runner environment file for instance {{ runner_instance }}
+  copy:
+    dest: "{{ runner_install_dir }}/{{ runner_instance }}/.env"
+    content: |
+      COMPOSE_PROJECT_NAME={{ runner_project_prefix }}-{{ runner_instance }}
+      INFINITO_RUNNER_PREFIX={{ runner_project_prefix }}-{{ runner_instance }}
+      SUBNET=172.30.{{ runner_instance }}.0/24
```
kevinveenbirkenbach: comment why

roles/svc-runner/tasks/instance.yml
```diff
@@ -0,0 +1,97 @@
...
+      INFINITO_CPUS={{ runner_cpus }}
+    owner: "{{ runner_user }}"
+    group: "{{ runner_user }}"
+    mode: '0600'
+
+- name: Install GitHub Actions runner service for instance {{ runner_instance }}
+  command: ./svc.sh install {{ runner_user }}
+  args:
+    chdir: "{{ runner_install_dir }}/{{ runner_instance }}"
+  when: not DOCKER_IN_CONTAINER
```
kevinveenbirkenbach: This doesn't seem right. validate. it or comment it.

roles/svc-runner/tasks/main.yml
```diff
@@ -0,0 +1,150 @@
...
+    append: true
+    system: true
+    shell: /usr/sbin/nologin
+    home: "{{ runner_install_dir }}"
+    create_home: true
+
+- name: Grant passwordless sudo for CI cleanup operations
+  copy:
+    dest: "/etc/sudoers.d/{{ runner_user }}"
+    content: |
```
kevinveenbirkenbach: use a extra conf file in sudoers.d/ <- It's cleaner
Also use an extra .j2 file and place it there to keep it cleaner.

.github/workflows/test-deploy-universal.yml
AlejandroRomanIbanez: These are for send CI to the runners, we already have the test for that in there, this should be here if not the self hosted logic serve no purpose

.github/workflows/test-deploy-workstation.yml
AlejandroRomanIbanez: These are for send CI to the runners, we already have the test for that in there, this should be here if not the self hosted logic serve no purpose

.github/workflows/test-code-lint.yml
```diff
@@ -34,5 +34,7 @@ jobs:
       - name: Build step summary from annotations
         if: always()
         working-directory: ${{ github.workspace }}
```
kevinveenbirkenbach, line 36: why?

.github/workflows/test-deploy-server.yml
```diff
@@ -47,10 +47,6 @@ jobs:
         shell: bash
         run: ./scripts/install/apt.sh jq
 
-      - name: Install deps (jq + envsubst)
```
kevinveenbirkenbach, line 52: If this works in CI/CD over all runs then fine. propably death code.

roles/svc-runner/files/test.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
+# Verifies svc-runner by deploying a real app inside the runner's own Docker
+# volume and project namespace. Variables are sourced from test.env.j2 by test-e2e-cli.
+set -euo pipefail
+
+: "${RUNNER_INSTALL_DIR:?}"
+: "${RUNNER_USER:?}"
+: "${DOCKER_IN_CONTAINER:?}"
```
kevinveenbirkenbach: Isn't it everytime DiD when called? Not sure.

roles/svc-runner/templates/test.env.j2
kevinveenbirkenbach, line 1: Where is the runner configured.

roles/svc-runner/files/test.sh
kevinveenbirkenbach, line 1: Filter output of make command if runner is really triggered on localhost. search for string fail if not existing.

.github/workflows/test-deploy-universal.yml
AlejandroRomanIbanez: templates/test.env.j2 + files/test.sh exist

.github/workflows/test-deploy-workstation.yml
AlejandroRomanIbanez: templates/test.env.j2 + files/test.sh exist

roles/svc-runner/files/test.sh
AlejandroRomanIbanez: Added the missing non-DinD block, it runs compose ps, filters the output and fails if fail/exit/dead appears. I leave the DOCKER_IN_CONTAINER since now is needed

roles/svc-runner/templates/test.env.j2
AlejandroRomanIbanez: Runner is configured in defaults/main.yml (install dir, user, count, etc.) and vars/main.yml (GitHub owner/repo/token, sourced from env vars at deploy time). Operator overrides go via make runner-ci-deploy flags or environment variables.

See README → Role variables

cli/administration/deploy/development/down.py
```diff
@@ -2,65 +2,54 @@
...
-CI_DOCKER_ROOT = Path("/mnt/docker")
-CI_DOCKER_ROOT_STR = str(CI_DOCKER_ROOT)
 
+def _resolve_docker_root() -> Path:
+    raw = (os.environ.get("INFINITO_DOCKER_VOLUME") or "").strip().rstrip("/")
+    if not raw:
+        raise RuntimeError("INFINITO_DOCKER_VOLUME must be set")
+    return Path(raw)
 
-def _base_env(*, distro: str) -> dict[str, str]:
```
kevinveenbirkenbach, line 19: Not sure if this deletions make sense

docs/contributing/tools/agents/cheatsheet.md
```diff
@@ -44,6 +45,14 @@ Follow the instructions from AGENTS.md, iterate on web app role <role> by follow
 
 SPOT: [Role Loop](../../../agents/action/iteration/role.md)
 
+## Runner Iteration 🏃
```
kevinveenbirkenbach: This should not be necessary because iteration/role.md should behave the same on runners like on local deploys.

roles/svc-runner/defaults/main.yml
kevinveenbirkenbach, line 1: Should be in vars/main.yml. And everything as CONSTANT (Upercase). confifgurable options deifend in meta/services.yml and read it out vars/main.yml.

roles/svc-runner/defaults/main.yml
```diff
@@ -0,0 +1,19 @@
+runner_name: "{{ inventory_hostname }}"
+runner_distribution: ""
+runner_labels: "self-hosted,linux,{{ runner_distribution }}"
+runner_install_dir: /opt/github-runner
+runner_user: github-runner
+runner_uid: 900
```
kevinveenbirkenbach: define user in meta/users.yml, to have one spot and guaranty that GID and UID don't overlapse with other users.

roles/svc-runner/defaults/main.yml
```diff
@@ -0,0 +1,19 @@
...
+runner_gid: 900
+# Matches the GitHub-hosted 2-core runner quota; drives auto-scaling via runner_count.
+runner_cpus: 2
+runner_count: "{{ [ansible_processor_vcpus | int // runner_cpus | int, 1] | max }}"
+runner_docker_base: "{{ lookup('config', application_id, 'services.svc-runner.docker_base') }}"
+runner_project_prefix: runner
+runner_sysctl_conf: /etc/sysctl.d/99-github-runner.conf
+runner_image_name: infinito-runner
+runner_image_tag: latest
+# GitHub registration — lowest precedence so -e extra-vars always win
```
kevinveenbirkenbach: Default values early in the chain best in meta/services.yml

roles/svc-runner/files/Dockerfile
```diff
@@ -0,0 +1,44 @@
+FROM ubuntu:22.04
+
+ARG RUNNER_UID=900
```
kevinveenbirkenbach: Defined earlier in code as constant, pass this as env and use it here. everytime just one SPOT.

roles/svc-runner/files/Dockerfile
```diff
@@ -0,0 +1,44 @@
+FROM ubuntu:22.04
```
kevinveenbirkenbach: Don't hardcode image otherwise mirror doens't store them.

roles/svc-runner/files/test/external.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
```
kevinveenbirkenbach, line 1: execute make reinstall dashboard with all services disabled. 

See https://github.com/infinito-nexus/core/tree/main/scripts/tests/environment

Like the envioronment test is going through the full environment setup. This should really test a real life scenario how to runners are woirking to verify e2e that everything runs accordingly.

roles/svc-runner/meta/services.yml
kevinveenbirkenbach, line 1: Not sure if this is the right spot recheck it :)

roles/svc-runner/tasks/install/packages.yml
```diff
@@ -0,0 +1,45 @@
+- name: Update apt package cache (Debian/Ubuntu)
```
kevinveenbirkenbach: Not sure if this fixes the root cause. May try to delte all images local rebuild and may cashing issue is solved.

roles/svc-runner/tasks/main.yml
```diff
@@ -0,0 +1,223 @@
+- name: Mark swapfile as already handled — runner manages its own memory via Docker limits
```
kevinveenbirkenbach: I think this could lead to the deactivation of swapfile if it's deployed with other apps together and that could be an issue for users expectign swapfile to work.

roles/svc-runner/tasks/main.yml
```diff
@@ -0,0 +1,223 @@
...
+- name: Set docker GID fact
+  ansible.builtin.set_fact:
+    _runner_docker_gid: "{{ (_docker_group_entry.stdout | split(':'))[2] | default('999') }}"
+
+- name: Install compose wrapper
+  ansible.builtin.include_role:
+    name: sys-svc-compose
+    tasks_from: 01_core.yml
+
+- name: Render runner compose file
```
kevinveenbirkenbach: Not 100% sure but compose.yml.j2 should be auto copied by sys-svc-compose

cli/administration/deploy/development/down.py
AlejandroRomanIbanez: it was a helper function for calling docer compose with a subprocess, it has been replaced by Compose class now, is a refactor rather than a full deletion

roles/svc-runner/tasks/main.yml
AlejandroRomanIbanez: WIll be the case if we call it from via main.yml, but in this case I called from 01-core.yml only, this is to aovid destination mistmatches with co-located runner binary

roles/svc-runner/files/test.sh
```diff
@@ -0,0 +1,184 @@
...
+: "${RUNNER_GITHUB_OWNER:?}"
+: "${RUNNER_GITHUB_REPO:?}"
+: "${RUNNER_INSTALL_DIR:?}"
+: "${RUNNER_PROJECT_PREFIX:?}"
+: "${RUNNER_USER:?}"
+
+echo "OK: env vars verified (RUNNER_COUNT=${RUNNER_COUNT}, RUNNER_PROJECT_PREFIX=${RUNNER_PROJECT_PREFIX}, RUNNER_GIT_REF=${RUNNER_GIT_REF})"
+
+# --- Binary check (skip gracefully if svc-runner was not deployed in this run) ---
+if [[ ! -f "${RUNNER_INSTALL_DIR}/1/run.sh" ]]; then
```
kevinveenbirkenbach: This shouldn't be done. discover mechanism should  just call files which are actually deployed.

roles/svc-runner/files/test.sh
```diff
@@ -0,0 +1,184 @@
...
+    echo "SKIP: svc-runner not deployed in this run (binary absent at ${RUNNER_INSTALL_DIR}/1/run.sh)"
+    exit 0
+fi
+echo "OK: runner binary present at ${RUNNER_INSTALL_DIR}/1/run.sh"
+
+fail_count=0
+
+# --- DinD: registration and Docker socket are skipped (entrypoint sleeps); only verify containers started ---
+if [[ "${DOCKER_IN_CONTAINER}" == "true" ]]; then
```
kevinveenbirkenbach: test-e2e-cli role should use same discovery mechanism like playwright test role. As far as I remember it sjust triggers in general when DOCKER:_IN_CONTAINER.

roles/svc-runner/files/test/external.sh
```diff
@@ -0,0 +1,184 @@
...
+# E2E verification for svc-runner.
+# Checks: binary presence, container health, Docker socket access from inside each
+# container (DooD — proves runner can execute CI jobs), GitHub registration, and
+# dispatches test-runner-smoke.yml to verify a real end-to-end CI job runs to success.
+# Variables sourced from test.env.j2 by test-e2e-cli. RUNNER_API_TOKEN from env only.
+set -euo pipefail
```
kevinveenbirkenbach, line 5: Verify that they fail ahrd when command fails.

roles/svc-runner/files/test/external.sh
```diff
@@ -0,0 +1,184 @@
...
+    # Proves the runner can actually execute CI job steps that use Docker.
+    if ! docker exec "${container}" docker version >/dev/null 2>&1; then
+        echo "FAIL: Docker socket not accessible from inside ${container} — DooD broken; runner cannot run CI jobs"
+        fail_count=$((fail_count + 1))
+    else
+        echo "OK: ${container} can reach Docker socket (DooD works)"
+    fi
+done
+
+# 4) GitHub API: verify at least RUNNER_COUNT runners are registered and online.
```
kevinveenbirkenbach, line 14: if possible split up between:

test/test.sh -> orchestrator
test/local.sh
test/external.sh

roles/svc-runner/files/test.sh
AlejandroRomanIbanez: Fixed. discover_cli_roles now accepts group_names and filters on it, matching the behavior of application_allowed during the deploy phase.

The stage trigger was already configured as RUNTIME in ['dev', 'act', 'github'], which is identical to the Playwright implementation.

The DOCKER_IN_CONTAINER check at line 29 remains unchanged because it is not a trigger; it is a behavioral switch. svc-runner is CI infrastructure, so when running in DinD there is no real GitHub token available. Allowing runners to register from within a CI job would be unsafe, as they could potentially pick up and execute real workflows.

Instead, the entrypoint sleeps to keep the container alive without attempting registration. Within DinD we test everything that is realistically possible:

Containers start successfully
The runner binary is present (verified by the entrypoint before sleeping)
The DooD socket is mounted correctly

The full end-to-end flow (runner registration and smoke dispatch) is executed on a real host using RUNNER_API_TOKEN. This follows the same principle as the Playwright setup, except the external GitHub dependency cannot be fully exercised from within CI.

roles/svc-runner/files/test/local.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
+# Local checks for svc-runner: container health, runner process, and DooD.
+# In DinD mode the runner entrypoint sleeps (no GitHub token available);
+# only container start is verifiable in that environment.
+set -euo pipefail
+
+fail_count=0
+
+# --- DinD: entrypoint sleeps; only verify containers started ---
+if [[ "${DOCKER_IN_CONTAINER}" == "true" ]]; then
```
kevinveenbirkenbach: Redundant if -> Tests anyhow are just called in DiD environment.

roles/svc-runner/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
```
kevinveenbirkenbach, line 1: UPPERCASE for constants.

roles/svc-runner/vars/main.yml
kevinveenbirkenbach, line 1: UPPERCASE for constants.

roles/svc-runner/files/test/local.sh
kevinveenbirkenbach, line 1: Please put output log in the PR.

kevinveenbirkenbach, line 1: Image

https://cloud.infinito.nexus/apps/files/files/265444?dir=/Infinito.Nexus/IT/Architecture/CORE&editing=false&openfile=true

roles/svc-runner/tasks/02_cleanup_instance.yml
kevinveenbirkenbach, line 1: XX_cleanup_instance.yml

roles/svc-runner/tasks/03_instance.yml
kevinveenbirkenbach, line 1: XX_instance.yml

roles/svc-runner/tasks/main.yml
kevinveenbirkenbach, line 1: 01_core.yml with serviceguard in main.yml

roles/svc-runner/templates/runner_instance_env.j2
```diff
@@ -0,0 +1,25 @@
...
+INFINITO_SUBNET=172.30.{{ runner_instance }}.0/24
+INFINITO_GATEWAY=172.30.{{ runner_instance }}.1
+INFINITO_DNS_IP=172.30.{{ runner_instance }}.53
+INFINITO_IP4=172.30.{{ runner_instance }}.10
+INFINITO_BIND_IP=172.30.{{ runner_instance }}.1
+# Cache static IPs must move with the per-instance subnet too, else compose
+# rejects them with "no configured subnet contains IP address".
+INFINITO_CACHE_PACKAGE_FRONTEND_IP=172.30.{{ runner_instance }}.4
+INFINITO_CACHE_REGISTRY_IP=172.30.{{ runner_instance }}.6
+DOMAIN=infinito.example
```
kevinveenbirkenbach: don't use fixed domains, use variables from group_vars, could also be infinito.local or something else

roles/svc-runner/templates/compose.yml.j2
```diff
@@ -0,0 +1,21 @@
...
+    image: "{{ RUNNER_IMAGE_NAME }}:{{ RUNNER_IMAGE_TAG }}"
+    container_name: runner-{{ i }}
+    restart: unless-stopped
+    working_dir: "{{ RUNNER_INSTALL_DIR }}/{{ i }}"
+    env_file: "{{ RUNNER_INSTALL_DIR }}/{{ i }}/.env"
+    volumes:
+      - "{{ RUNNER_INSTALL_DIR }}/{{ i }}:{{ RUNNER_INSTALL_DIR }}/{{ i }}"
+      - /var/run/docker.sock:/var/run/docker.sock
+      - "{{ RUNNER_DOCKER_BASE }}/{{ i }}:{{ RUNNER_DOCKER_BASE }}/{{ i }}"
+{% if DOCKER_IN_CONTAINER | bool %}
```
kevinveenbirkenbach: use container cp, in the local.sh, to keep it in one spot for testing.

roles/svc-runner/files/test/local.sh
```diff
@@ -0,0 +1,85 @@
...
+    DOCKER_HOST= docker save "$img" | docker load
+    COMPOSE_PROJECT_NAME=infinito INFINITO_RUNNER_PREFIX=infinito \
+    RUNTIME=github CI=true apps=web-app-dashboard disable=matomo \
+    INFINITO_DEPLOY_TYPE=server INFINITO_DISTROS=debian \
+    INFINITO_IMAGE="$img" INFINITO_BUILD=0 INFINITO_PULL_POLICY=never \
+    GITHUB_REPOSITORY_OWNER= GITHUB_REPOSITORY= \
+    INFINITO_INVENTORY_DIR=/tmp/runner-dind-inventory \
+    INFINITO_DOCKER_VOLUME=/tmp/runner-dind-docker \
+    ANSIBLE_LOG_PATH=/tmp/ansible-runner-dind-test.log \
+    bash scripts/tests/deploy/ci/all.sh
```
kevinveenbirkenbach: make deploy dashboard with all services disabled is missing

kevinveenbirkenbach (commented):
Seems overall good. Please change the last things. when the CI passed and the last comments are changed I'm going to merge it.

kevinveenbirkenbach:
Changed status to ready for review to have a final run here on origin. Needs to pass here before it can be merged.

.github/workflows/test-runner-smoke.yml
kevinveenbirkenbach, line 1: Write security implications of hacking due to executing PR's of runner servers which break out of docker.

compose/package-cache-frontend/upstreams.conf
kevinveenbirkenbach, line 1: http://package-cache:8081 --> everytime when you use a string twice set a SPOT.

roles/svc-prx-openresty/tasks/01_core.yml
kevinveenbirkenbach, line 1: Seems like a hack but ho show stopper. If somebody< can refactor it in the future would nbe nice.

roles/svc-runner/files/test/local.sh
```diff
@@ -0,0 +1,67 @@
...
+
+echo "DinD mode: verifying DooD socket in ${RUNNER_PROJECT_PREFIX}-1..."
+if ! docker exec "${RUNNER_PROJECT_PREFIX}-1" docker version >/dev/null 2>&1; then
+    echo "FAIL: Docker socket not accessible inside ${RUNNER_PROJECT_PREFIX}-1 — DooD broken in DinD"
+    exit 1
+fi
+echo "OK: DooD socket accessible inside ${RUNNER_PROJECT_PREFIX}-1"
+
+# The full deploy is identical across the sync/async passes; run it once (sync).
+if [[ "${ASYNC_ENABLED:-false}" == "true" ]]; then
```
kevinveenbirkenbach, line 29:
```suggestion
if [[ "${ASYNC_ENABLED}" == "true" ]]; then
```

roles/svc-runner/files/test/local.sh
```diff
@@ -0,0 +1,67 @@
...
+    echo "ALL LOCAL CHECKS PASSED (DinD)"
+    exit 0
+fi
+
+# Full reinstall deploy of the dashboard inside ephemeral runner-1. No GitHub/GHCR.
+_iso_src="${RUNNER_INSTALL_DIR}/1/nested-src"
+echo "DinD mode: running full reinstall deploy inside ${RUNNER_PROJECT_PREFIX}-1..."
+# container cp the repo into runner-1 (no compose mount — keeps test wiring here),
+# then isolate a per-instance copy (drop .env/Corefile so inner coredns serves the right IP).
+container exec --user root "${RUNNER_PROJECT_PREFIX}-1" mkdir -p /opt/src/infinito
```
kevinveenbirkenbach, line 40: Next time use spot for /opt/src/infinito. Passing this time.

.github/workflows/test-runner-smoke.yml
kevinveenbirkenbach, line 1: No extra runner should be required. Should be sufficient to pass config viar test.env.j2 to test.sh which calls external.sh. No show stopper,  but refactoring needed in futre.

roles/svc-runner/files/Dockerfile
kevinveenbirkenbach, line 1: In documentation warn about runners on non isolated servers.

roles/sys-svc-compose/templates/networks.yml.j2
```diff
@@ -54,7 +54,7 @@ networks:
 {% if not application_id.startswith('svc-db-') and not application_id.startswith('svc-ai-') %}
   default:
 {% if _default_subnet %}
-    name: {{ application_id | get_entity_name }}
+    name: {{ application_id | get_entity_name or application_id }}
```
kevinveenbirkenbach, line 57: Don't thing this is necessary. Anyhow ius soon as I merge it to swarm will be anyhow refactored.

scripts/tests/deploy/ci/all.sh
```diff
@@ -144,8 +144,12 @@ for distro in "${distro_arr[@]}"; do
 	distro_start="$(date +%s)"
 
 	set +e
```
kevinveenbirkenbach, line 146: Theoretically can be deleted.

PR #403, opened 2026-07-09
feat(web-app-stalwart): Stalwart as default mail provider (Mailu opt-in) + .test dev domain

AlejandroRomanIbanez:
## Summary

Make [Stalwart](https://stalw.art) (`stalwartlabs/stalwart:v0.16.12`) the platform's default mail provider with full Mailu feature parity: SMTP/IMAP/POP3/ManageSieve, CalDAV/CardDAV, Roundcube webmail, Keycloak OIDC SSO, DKIM/DNS provisioning, ClamAV milter, `no-reply` token persistence via `sys-token-store`, and role-local Playwright e2e. Mailu stays fully supported as a one-line opt-in (`MAIL_PROVIDER: web-app-mailu`) and can coexist on the same host in legacy mode.

Because Stalwart's compiled-in validation rejects RFC-6761-reserved domains, the local/CI development domain moves `infinito.example` → **`infinito.test`** (single SPOT in `default.env`); a fail-fast assert in the role replaces all previous reserved-TLD workaround gating.

---

## Template Type

* [x] **Feature** - Adds or extends server functionality
* [ ] **Fix** - Repairs broken or incorrect server behavior

---

* Primary `web-*` role(s): **`web-app-stalwart`** (default provider), **`web-app-mailu`** (legacy-mode coexistence)
* Related role(s): `sys-svc-mail` (provider-agnostic preload assert), `web-app-keycloak` (WebAdmin + webmail OIDC clients), `sys-service-loader` (deferred-restart post-load queue), `sys-front-tls-selfsigned`, `sys-svc-compose-ca`, `web-app-prometheus` (Alertmanager `smtp_hello`), email-flag migration in `web-app-checkmk` / `web-app-n8n` / `web-app-semaphore` / `web-app-joomla`, dev DNS/TLS scripts for `.test`.

## Preferred Integrations

* [x] Dashboard
* [x] Matomo
* [x] OIDC
* [ ] LDAP — Stalwart's auth directory is internal XOR **one** external directory; OIDC (Keycloak, itself LDAP-federated) was chosen so web SSO works. A direct-LDAP variant remains possible (see Additional Notes).
* [x] Logout

---

```yaml
roles:
  - web-app-stalwart
  - web-app-mailu
```

---

## Change Type

* [x] **Major** - Breaking change
* [ ] **Minor** - New backwards-compatible feature

**Breaking:** inventories that deploy Mailu without setting `MAIL_PROVIDER` now get Stalwart on `mail.<domain>`; Mailu moves to `legacy-mail.<domain>` and stops publishing the public mail ports. Set `MAIL_PROVIDER: web-app-mailu` to keep today's behavior exactly.

---

## Change Details

* **Problem solved:** replaces the multi-container Mailu stack with a single hardened mail server that integrates natively with the platform's OIDC SSO, while keeping Mailu selectable.
* **Provider selector:** `MAIL_PROVIDER` (group_vars) routes `lookup('email')` and every consumer `services.email` flag; consumers migrated off hardcoded `'web-app-mailu' in group_names`.
* **Port ownership:** only the active `MAIL_PROVIDER` publishes 25/465/587/110/995/143/993/4190 — both providers coexist in one play (CI matrix rounds require this); mirrors Mailu's existing `legacy-mail.` vhost fallback.
* **OIDC ordering (both providers):** the mail provider must preload before Keycloak (`sys-svc-mail` asserts its `no-reply` token), but both Stalwart (issuer resolved at restart, failure sticks) and heviat-Mailu (provider config cached from the first `/sso` hit) need Keycloak *up* to activate SSO. Solved with a restart-or-defer pattern through the `sys-service-loader` post-load queue: probe the issuer from the app container; reachable → restart now; not yet deployed → queue the role for the post-load flush.
* **WebAdmin staging:** the WebUI bundle (`stalwartlabs/webui v1.0.5`, pinned) is fetched by the controller into a host cache (`/var/cache/infinito/`, survives inter-round purges — per-round GitHub re-downloads hit runner rate limits), mounted read-only, and unpacked via the JMAP `UpdateApps` action (the legacy `/api/update/webadmin` endpoint is gone in v0.16); a loud post-unpack probe fails the deploy on a 404 WebAdmin.
* **Credential convergence:** account provisioning now updates the password of pre-existing accounts (the `primaryKeyViolation` response carries the id) — a create-only flow left the DB on stale secrets after reuse/update deploys (`hlth-msmtp` 535).
* **Mailu legacy mode:** `HOSTNAMES` + vhost now use the role's own canonical (`MAILU_CANONICAL_HOST`) instead of `lookup('email').host` — the old value pointed at the *active provider's* vhost, which both broke heviat's OIDC host allow-list and let legacy Mailu overwrite Stalwart's `mail.` vhost. Legacy instances run `TLS_FLAVOR=notls` behind the proxy and skip `sys-ctl-mtn-cert-deploy` (shared run-once unit — belongs to the active provider).
* **Pinned versions:** Stalwart `v0.16.12`, Roundcube `1.7.2-apache`, ClamAV `1.5.3`, WebUI `v1.0.5` — mutable `latest` previously made CI (mirror) and local hosts run different builds with different HTTP surfaces.
* **`.test` switch:** `default.env` is the SPOT; env-test bootstrap falls back to `default.env` when python3 isn't installed yet; dev DNS routes the zone via a systemd-resolved drop-in (per-link DNS on loopback is always rejected).
* **Drive-by fixes found via the CI matrix:** `sys-svc-compose-ca` no longer pulls locally-**built** image tags (the registry base image's CMD got pinned into the CA override → crash-loop, e.g. bridgy-fed); `sys-front-tls-selfsigned` SAN regex had `\\s` (literal `s` in the class → every SAN truncated → cert regenerated each deploy); Alertmanager gets `smtp_hello` (Stalwart rejects a bare container hostname EHLO with `550`); the project-tree walker honors `.git/info/exclude`.
* **Alternatives considered:** keeping Mailu as default (rejected — Stalwart is the strategic choice); keeping `.example` with per-app workarounds (rejected by maintainer — fail-fast assert + `.test`); delete-and-recreate token flow like Mailu's (Stalwart's JMAP update is cheaper and idempotent).

---

## File Checklist

| Check | Item |
|---|---|
| [x] | `README.md` (incl. the internal-XOR-external directory rationale) |
| [x] | `meta/main.yml` |
| [x] | `vars/main.yml` |
| [x] | `meta/services.yml` (pinned versions, `provides: email`, ports, `run_after`) |
| [x] | `meta/server.yml` (`mail.` / `webmail.` canonicals, CSP connect-src for the SPA's OIDC code exchange, subnet) |
| [x] | `meta/rbac.yml` (`bot`) |
| [ ] | `meta/volumes.yml` — volumes declared inline via `compose_volumes` |
| [x] | `meta/schema.yml` (admin credential) |
| [x] | `tasks/main.yml` (+ `01`–`12` task files incl. deferred OIDC restart) |
| [x] | `templates/compose.yml.j2` (stalwart + webmail + clamav; provider-gated public ports) |
| [ ] | `templates/env.j2` — Stalwart bootstraps from `config.json.j2` (data-store config) |
| [ ] | `style.css` / `javascript.js` — upstream WebAdmin SPA + stock Roundcube |
| [x] | `meta/users.yml` |
| [ ] | `files/Dockerfile` — upstream images, pinned |
| [x] | `templates/playwright.env.j2` |
| [x] | `files/playwright/playwright.spec.js` (WebAdmin TLS, CalDAV/CardDAV discovery, SSO login/logout, send→receive, guest denial) |
| [x] | `meta/variants.yml` (all-on / all-off rounds) |

### Registered

| Check | Item |
|---|---|
| [x] | Ports in `meta/services.yml` (`stalwart` local http 8081, `webmail` 8089 — via the suggest CLIs after upstream collisions) |
| [x] | Subnet in `meta/server.yml` (`networks.local.subnet: 192.168.106.32/28`) |

---

## Local Validation

* [x] `make test` green — lint **157**, integration **196**, unit **2664**, external.
* [x] Deployment target documented — DinD dev stack, debian; both matrix variants, plus `mode=update` redeploys over kept state (the scenario that kept failing in CI) for both providers.
* [x] Playwright documented — Stalwart 5/5 per round; Mailu legacy `sso login … logout` ✓ (13.8s) and `biber sends → administrator receives` ✓ (16.2s).
* [x] Login/logout tested — Keycloak SSO into WebAdmin and Roundcube (both providers); universal logout specs.
* Every CI-observed failure class was reproduced and re-verified locally before its fix landed, e.g.:

```
make compose-deploy apps=web-app-stalwart               # both rounds
make compose-deploy mode=update apps=web-app-stalwart variant=1
make compose-deploy apps=web-app-mailu                  # legacy coexistence
make compose-playwright role=web-app-stalwart
```

---

## Security Impact

* [x] Security impact present

* **Surfaces:** public mail ports now published **only** by the active provider (smaller exposure when both are installed); with SSO enabled Stalwart switches all interactive auth to the Keycloak OIDC directory and disables password login; `no-reply` submission credential lives in `sys-token-store`; trusted-network relay restricted to internal subnets.
* **Risk reduction:** image pinning removes mutable-`latest` supply drift; the SAN parse fix stops needless daily cert regeneration.
* **Validation:** guest-never-authenticated denial spec; `sys-svc-mail` hard-asserts provider preload + token presence (it caught a real regression during development).

---

## Review Focus

* Highest-risk: the `application_id == MAIL_PROVIDER` port gating in both compose templates; the restart-or-defer OIDC pattern (`web-app-stalwart/tasks/12_oidc_apply.yml`, `web-app-mailu/tasks/09_oidc_apply.yml`, service-loader queue re-entry in both `main.yml`); the `MAIL_PROVIDER` default flip.
* Migration: existing Mailu deployments must set `MAIL_PROVIDER: web-app-mailu` (see Change Type).
* Known-not-this-PR: `espocrm`, `fider`, `n8n` deploy jobs fail identically on current `main`; `gitlab` on arch (and `nextcloud [5,6]`, which provisions against GitLab) look like runner-load, pending one clean run to confirm.

---

## Definition of Done (DoD)

---

## Additional Notes

* **Draft during the Swarm freeze.** Compose is the base per the freeze announcement; after Swarm lands this role needs a compatibility pass, and the public-port gating is a known swarm-mode discussion point.
* **Production impact is zero-config:** the mail domain derives from the inventory's primary domain; `.test` is dev/CI-only.
* **LDAP variant (future work, decision for the maintainer):** Stalwart supports an LDAP directory type, but directory choice is exclusive — a `services.ldap` variant could bind straight to OpenLDAP (native mail-client password auth) at the cost of web SSO, analogous to Moodle's LDAP-only variant. Not needed for Mailu parity (Mailu had no LDAP flag either).
* **Latent upstream gap observed (not addressed here):** `sys-ctl-mtn-cert-deploy` is a global run-once unit — only one app per host can use it. Legacy Mailu sidesteps it; a second cert-consuming app on one host would hit it.


cli/meta/domains/__main__.py
kevinveenbirkenbach, line 1: MAY it helps you to write a regression test when swarm will be merged into this.

group_vars/all/00_general.yml
```diff
@@ -55,6 +55,8 @@ DNS_PROVIDER: "localhost" # The DNS Provider\Registrar for the domain -> Legit v
 
 HOSTING_PROVIDER: "localhost" # Provider which hosts the server -> hetzner
 
+MAIL_PROVIDER: "web-app-stalwart" # Active mail provider role. Set to web-app-mailu to use Mailu instead of Stalwart.
```
kevinveenbirkenbach, line 58: make dependend on in group_names -> fail hard if mailu and stalwart are deployed, just one provider possible.

group_vars/all/14_about.yml
```diff
@@ -20,7 +20,7 @@ defaults_service_provider:
     web-app-bluesky: >-
       {{ ('@' ~ lookup('users', 'contact').username ~ '.' ~ lookup('domains', 'web-app-bluesky').api)
          if 'web-app-bluesky' in group_names else '' }}
-    email: "{{ lookup('users', 'contact').username ~ '@' ~ DOMAIN_PRIMARY if 'web-app-mailu' in group_names else 'contact@infinito.nexus' }}"
+    email: "{{ lookup('users', 'contact').username ~ '@' ~ DOMAIN_PRIMARY if 'web-app-stalwart' in group_names else 'contact@infinito.nexus' }}"
```
kevinveenbirkenbach: Not sure if MAIL_PROVIDER or in [ ] is right approach

group_vars/all/00_general.yml
kevinveenbirkenbach, line 58: There is a lookup plugin, may it makes sense to place it there. not 100% sure. think name is email

roles/sys-ca-selfsigned/files/with-ca-trust.sh
```diff
@@ -37,11 +37,26 @@ log "Sanitized trust name: $name"
...
-export SSL_CERT_FILE="$CA_TRUST_CERT"
-export REQUESTS_CA_BUNDLE="$CA_TRUST_CERT"
-export CURL_CA_BUNDLE="$CA_TRUST_CERT"
-# Optional (harmless if unused; helps Node-based tools if any)
+# Env-based trust fallback. Build a COMBINED bundle (system CAs + our CA) so the
+# env vars don't break public-HTTPS validation; fall back to our CA only if no
+# system bundle is found.
+ca_bundle="$CA_TRUST_CERT"
+combined="/tmp/with-ca-trust-combined.crt"
+for sys_bundle in \
```
kevinveenbirkenbach: Refactor. To SPOT used at multiple ocassions.

roles/sys-ctl-mtn-cert-deploy/meta/services.yml
kevinveenbirkenbach, line 1: Integrate test.sh which tests local and additonal test a server which is passed over env if it is right configured. 

Idea behind this: 
e.g. infinito.nexus or lajeandros.server give failure when wrong congfigured and we see it instantly in the CI. 

Should be configured via github variables \ env file.

kevinveenbirkenbach: Integrate prometheus role if not existing.

group_vars/all/00_general.yml
AlejandroRomanIbanez: The lookup already carries the same default as a fallback (DEFAULT_MAIL_PROVIDER in plugins/lookup/email.py) and prefers the var. The group_var itself has to stay: meta/services.yml flags template MAIL_PROVIDER in group_names as an Ansible var

Ill check this again after Swarm cahnges

roles/sys-ctl-mtn-cert-deploy/meta/services.yml
AlejandroRomanIbanez: the existing blackbox https probes already export probe_ssl_earliest_cert_expiry, so no new exporter was needed: added TlsCertificateExpiringSoon (<14d, warning) and TlsCertificateExpiryCritical (<3d, critical) alert rules

roles/web-app-stalwart/README.md
kevinveenbirkenbach, line 1: Add mermaide

roles/sys-ca-selfsigned/files/with-ca-trust.sh
```diff
@@ -6,6 +6,10 @@ set -eu
 
 VERBOSE="${VERBOSE:-1}"
 
+# SPOT: system CA bundle candidates, in preference order
+# (Debian/Ubuntu, RHEL/Fedora, Alpine/openssl default).
+SYS_CA_BUNDLE_CANDIDATES="/etc/ssl/certs/ca-certificates.crt /etc/pki/tls/certs/ca-bundle.crt /etc/ssl/cert.pem"
```
kevinveenbirkenbach, line 11: Verify if is best SPOT

roles/sys-front-inj-matomo/tasks/inject.yml
kevinveenbirkenbach, line 1: Not sure if this is right. Anyhow swarm will propably break it.

roles/web-app-fider/meta/services.yml
```diff
@@ -42,7 +42,7 @@ fider:
       http: 8067
   run_after:
     - web-app-keycloak
-    - web-app-mailu
```
kevinveenbirkenbach, line 45: keep run_after after swarm merge. Currently propably not possiblöe so you can avoid it.

kevinveenbirkenbach (commented):
Additional: Migration test is needed for mailu to stalwart; 
test.sh -> setup mailu stump with some data. (not full infinito.nexus deploy neede)
Seed data. 
Execute Migration Script 
Verify that migration worked.

roles/web-app-keycloak/tasks/update/01c_stalwart_webmail_client.yml
kevinveenbirkenbach, line 1: Explain more in detail WHY.

roles/web-app-keycloak/templates/import/clients/stalwart_webui.json.j2
kevinveenbirkenbach, line 1: It would be nice to change all poossible fields over the keyclaok interface. Check if this is possible. If then integrate it.

roles/sys-ca-selfsigned/files/with-ca-trust.sh
AlejandroRomanIbanez: the script is a standalone asset bind-mounted into containers by compose_ca.py; it cannot source repo files at runtime,

roles/sys-front-inj-matomo/tasks/inject.yml
AlejandroRomanIbanez: the Matomo guard exists because CI actually hit it: Matomo answered HTTP 200 with an HTML error body, site_check.json stayed undefined, and the play died later with a misleading error. The guard turns that into an immediate, diagnosable failure

roles/web-app-keycloak/tasks/update/01c_stalwart_webmail_client.yml
AlejandroRomanIbanez: update inlione comments with more info

roles/sys-ca-selfsigned/files/with-ca-trust.sh
```diff
@@ -6,6 +6,10 @@ set -eu
 
 VERBOSE="${VERBOSE:-1}"
```
kevinveenbirkenbach, line 7:
```suggestion
: "${VERBOSE:?VERBOSE must be set}"
```

roles/sys-ctl-alm-compose/meta/services.yml
kevinveenbirkenbach, line 1: Seems like a recursion. the email provider should also need the alarm? or? Actually when then it makes more sense at sys-ctl-alm-email

roles/sys-svc-mail/meta/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
 galaxy_info:
   author: Kevin Veen-Birkenbach
-  description: Central mail orchestration role for Infinito.Nexus, integrating Mailu, msmtp, and an optional local SMTP relay.
+  description: Central mail orchestration role for Infinito.Nexus, integrating Stalwart, msmtp, and an optional local SMTP relay.
```
kevinveenbirkenbach:
```suggestion
  description: Central mail orchestration role for Infinito.Nexus, integrating Email SaaS, msmtp, and an optional local SMTP relay.
```

roles/web-app-akaunting/meta/services.yml
```diff
@@ -55,7 +55,7 @@ akaunting:
   run_after:
     - web-app-matomo
     - web-app-keycloak
-    - web-app-mailu
```
kevinveenbirkenbach, line 58: can run after both. change it everywhere that it runs after both.

roles/web-app-keycloak/library/keycloak_kcadm_update.py
kevinveenbirkenbach, line 1: Why is this necessary.

roles/web-app-prometheus/templates/configuration/alert_rules.yml.j2
kevinveenbirkenbach, line 1: Where is it running? If just in stalwart also add it to mailu (if quick poossible) Otherwise mailu will anyhow be changed to lifecycle "maintained"

kevinveenbirkenbach (commented):
Write a migration script mailu to stalwart. 
Doesn't need to be called every time. 
Add test.sh with a migration procedure and test (write script for migration kevin needs it). Test should just be triggered when environment variable INFINITO_TEST_MIGRATION is set. Should be configurable via github variables.

roles/web-app-stalwart/files/playwright/playwright.spec.js
```diff
@@ -0,0 +1,219 @@
...
+// Baseline: WebAdmin is served on the canonical domain with TLS.
+test("stalwart: WebAdmin is served under canonical domain with TLS", async ({ page }) => {
+  const response = await page.goto(`${appBaseUrl}/`);
+  expect(response, "Expected Stalwart response").toBeTruthy();
+  expect(response.status(), "Expected status < 400").toBeLessThan(400);
+  expect(response.url().includes(canonicalDomain), `Expected canonical domain "${canonicalDomain}"`).toBe(true);
+  expect(response.headers()["strict-transport-security"], "Stalwart must emit HSTS").toBeTruthy();
+});
+
+// CalDAV/CardDAV auto-discovery is reachable through the proxy (.well-known -> /dav/*).
```
kevinveenbirkenbach, line 93: Think about how to connect cardav etc. automaticly with mailu per user account. Write a linkl bridge whatever if possible.

roles/web-app-stalwart/meta/main.yml
```diff
@@ -0,0 +1,33 @@
+---
+galaxy_info:
+  author: Kevin Veen-Birkenbach
```
kevinveenbirkenbach: take yyour credits.

roles/web-app-stalwart/meta/server.yml
```diff
@@ -0,0 +1,31 @@
...
+    - 301
+    - 200
+domains:
+  # Named canonicals (first is the primary); `webmail` is the Roundcube vhost.
+  canonical:
+    mail: "mail.{{ DOMAIN_PRIMARY }}"
+    webmail: "webmail.{{ DOMAIN_PRIMARY }}"
+  aliases: []
+csp:
+  flags:
```
kevinveenbirkenbach, line 16: Check if such open coinfigurations are really necessary. Could be to broad.

kevinveenbirkenbach: MAy an cli tool is helpful which itterates over the running instance and deactivates DiD the csp and activates it an sees what fails.

roles/web-app-stalwart/meta/services.yml
```diff
@@ -0,0 +1,98 @@
...
+  enabled: true # nocheck: dynamic-flag
+  shared: "{{ 'svc-db-postgres' in group_names }}"
+  cpus: "0.8"
+  mem_reservation: 1g
+  mem_limit: 2g
+  pids_limit: 512
+# nocheck: dynamic-flag — role-local sub-component, always deployed with the stack
+# nocheck: playwright-service-flag — internal sub-component
+webmail:
+  enabled: true
```
kevinveenbirkenbach, line 68: Set both to true check if you use them as guard

kevinveenbirkenbach:
INFINITO_TEST_MIGRATION dafault false. manual ci workflow should have a field to set it true. Besides this should be possible to override it via github variables.

AlejandroRomanIbanez:
I moved this logic to a new branch for Swarm integration #429 @kevinveenbirkenbach

PR #380, opened 2026-07-05
feat(svc-net-tor): Tor onion-service provider (full onion node)

kevinveenbirkenbach:
## Summary

Adds a `svc-net-tor` role that turns a NAT/CGNAT/edge node into a **full Tor v3 onion node**: the node's `DOMAIN_PRIMARY` becomes its `.onion`, so the whole stack (web, SSO, LDAP, CA) resolves onion domains and is reachable over Tor **without a public IP, port forwarding, VPS or DynDNS**.

Design doc: `docs/requirements/031-svc-net-tor-onion.md`.

## What's included

- **`utils/tor_onion.py`** — offline Tor v3 key mint (address + `HiddenServiceDir` key files), verified against a known key→address test vector.
- **`cli .../inventory/onion` (`init-env`)** — pre-mints the node onion, writes `INFINITO_DOMAIN`(+`_RE`) into `.env` and authoritative key files into `.onion-identity/hs` (idempotent). The onion is a **deploy-time input**; no in-deploy random minting.
- **`roles/svc-net-tor`** — Tor sidecar (Debian-built, `network_mode: host`), installs the minted keys, `torrc` with one node HiddenService + a SOCKS proxy restricted to loopback/DiD ranges. `files/test.sh` probes every served onion domain over the Tor SOCKS.
- **tls** — `.onion` → TLS off (onion v3 crypto is the transport auth; LE cannot sign `.onion`) + app-level flavor override.
- **harness/DNS** — coredns resolves the onion zone via `INFINITO_DOMAIN`; the clearnet DNS sanity check is skipped for onion; `nginx server_names_hash_bucket_size` raised to 128 for long onion `server_name`s.
- **`test-e2e-playwright`** — gated browser SOCKS proxy (`PLAYWRIGHT_PROXY`) + navigation/action timeouts so specs reach `.onion` over Tor and fail fast.

## Verification

- `make test` green (2666 integration + unit + all lints).
- Live full-onion deploy: **dashboard + SSO (Keycloak) + 9 apps reachable over Tor** (HTTP 200 through the node hidden service), confirmed via `roles/svc-net-tor/files/test.sh`.

## Follow-ups (tracked, not in this PR)

- Mailu / mail delivery over Tor (`hlth-msmtp` TLS chain on onion).
- SSH as a hidden service.
- `prometheus` returns 500 over onion (app-specific).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #463, opened 2026-08-06
Make Playwright specs actually run, and fix the roles that broke when they did

kevinveenbirkenbach:
## Summary

Two contract halves that the documentation already required but nothing enforced: a Playwright spec is only ever executed if its role also ships `templates/playwright.env.j2`, and a blocked persona must say why it is blocked. 21 of 76 `web-app-*` roles shipped a spec that had never run once, and 89 persona blocks across 46 roles carried no reason at all.

Both are now linted, both gaps are closed, and the three roles whose real defects surfaced while closing them are fixed: Magento never published its nginx vhost, Listmonk's settings were always one boot late, and Mastodon's SeaweedFS scenario contradicted its own persona contract. `web-app-socialhome` graduates from an empty scaffold to a role that actually deploys.

---

## Template Type

* [ ] **Feature** - Adds or extends server functionality
* [x] **Fix** - Repairs broken or incorrect server behavior

Both apply; `Fix` is the honest primary intent — the new lints exist to repair a class of silent non-coverage, not to add a capability.

---

* Primary `web-*` role(s): **68 roles touched.** Substantive: `web-app-socialhome` (new implementation), `web-app-magento`, `web-app-listmonk`, `web-app-mastodon`. New `templates/playwright.env.j2` in 21 roles: chess, confluence, espocrm, funkwhale, gitlab, jira, lam, listmonk, littlejs, magento, mastodon, mini-qr, mobilizon, navigator, pgadmin, phpldapadmin, phpmyadmin, roulette-wheel, shopware, socialhome, suitecrm. Persona rationale + README section in 46 further roles.
* Related role(s): `roles/test-e2e-playwright` (consumer of both contracts, unmodified), `web-svc-coturn` (newly wired by socialhome), `web-app-keycloak`, `web-app-seaweedfs`, `sys-ctl-hlth-webserver`.

## Preferred Integrations

* [x] Dashboard
* [x] Matomo
* [x] OIDC
* [ ] LDAP
* [x] Logout

socialhome integrates dashboard/matomo/logout/css/prometheus as usual. It deliberately does **not** use OIDC: an oauth2-proxy in front of it would redirect `POST /federation/inbox` and `/api/pairing/accept` to Keycloak and break the peer-to-peer federation the software exists for. Listmonk's OIDC becomes active for the first time (see below).

---

```yaml
roles:
  # - web-app-socialhome
  # - web-app-listmonk
  # - web-app-magento
```

Left commented. A subset run would hide exactly what this PR needs to learn — how the 17 newly-armed specs behave in a full matrix.

---

## Change Type

---

### The gap

`test_has_spec.py` makes a Playwright spec mandatory for every `web-app-*` role and calls a missing one a hard error, because the persona contract must never be silently absent. Nothing enforced the other half. The runner discovers roles by `rglob("templates/playwright.env.j2")` (`discover_playwright_roles.py:53`), so a spec without that file is staged nowhere and executed never.

Verified rather than inferred: running the discovery filter returns 61 roles and none of the 21 is among them; `make compose-playwright` refuses without a rendered `.env` from a prior deploy; and `git log --all` on those paths shows the file never existed on any branch. **15 of the 21 deploy in CI today, reporting green while their only functional verification never fires.**

The same shape one level down: blocking a persona removes the only end-to-end proof that a role's auth chain works, and `playwright.specs.js.md:146` has always required a one-line rationale per flag. 89 flags carried none.

### The roles that broke while closing it

* **web-app-magento** — `01_setup.yml` copied `nginx.conf.sample` inside the bootstrap shell, but `magento/project-community-edition` 2.4.8's dist archive is 2583 bytes holding only `composer.json`; the sample ships in `magento2-base` and lands only after `composer install`. `set -e` made that fatal and the block's `creates:` guard made it permanent, so `/` returned 403 forever. Also `placement: manager`, `--base-url` without the double slash, `--no-dev`, and the nginx restarts gated on an actual change.
* **web-app-listmonk** — every entry in `LISTMONK_SETTINGS` was one boot late. Listmonk reads settings into koanf once at startup and builds its media store from them; the role wrote them with a direct `UPDATE` after the only `notify: compose up` had already flushed. So `upload.provider=s3`, `security.oidc`, `smtp` and `app.root_url` were **never in effect** — uploads went to the container filesystem and `/admin/login` rendered no SSO control at all.
* **web-app-mastodon** — the personas ran unconditionally although `tootctl accounts create` never sets a password, and `test-seaweedfs.js` drove the same admin journey while gated only on `seaweedfs`.
* **web-app-socialhome** — was a two-line comment. It now deploys `ghcr.io/social-home-io/socialhome:2026.6.16` (**not** jaywink/socialhome — a different codebase: aiohttp + SQLite + Preact, Ed25519/WebRTC rather than ActivityPub). No Postgres, no Redis, no worker.

### Upstream versions

`ghcr.io/social-home-io/socialhome:2026.6.16` (new, pinned). `listmonk/listmonk:v6.2.0` and `magento/project-community-edition:2.4.8` unchanged; both were probed against the real image during this work.

### Login / logout / storage / mail after the change

Listmonk's OIDC works for the first time, via a container recreate plus an `appearance.public.custom_js` snippet that submits the OIDC form — because `cmd/auth.go renderLoginPage` hardcodes `PasswordEnabled`, so the native form cannot be hidden and the shared helper's single click would take the wrong control. Its S3 uploads reach the bucket for the first time. socialhome authenticates against its own user table with no OIDC client; both its personas are declared blocked, with reasons.

### Alternatives considered

Teaching the shared `clickOidcLoginLink` to follow a second hop was designed and costed — it would unblock listmonk, fider and odoo at once — and rejected for this PR: only five roles route through that helper live, and a blind second click can hit a native submit button on nine of them. The role-local `custom_js` solves listmonk without touching code 61 roles depend on. The design is recorded for a follow-up.

For socialhome, the pgadmin model (oauth2-proxy sidecar) was rejected for the federation reason above; a `javascript.js.j2` logout injector was rejected because the prometheus model points its button at `OIDC.CLIENT.LOGOUT_URL`, which does not exist for a role without OIDC.

---

## File Checklist

| Check | Item | When to include | Purpose |
|---|---|---|---|
| [x] | `README.md` | Usually | socialhome rewritten; `## Persona contract opt-outs` added to 43 roles. |
| [x] | `meta/main.yml` | Usually | socialhome: description pointed at the wrong upstream project. |
| [x] | `vars/main.yml` | Usually | socialhome (new), listmonk (`appearance.public.custom_js`), magento. |
| [x] | `meta/services.yml` | Usually | socialhome (new), magento (`placement: manager`, lifecycle), listmonk, mastodon, suitecrm, gitlab, shopware. |
| [ ] | `meta/server.yml` | Usually | Unchanged. listmonk's `status_codes.default: 404` was investigated and confirmed correct — see Review Focus. |
| [ ] | `meta/rbac.yml` | When the role declares RBAC | Not applicable. |
| [x] | `meta/volumes.yml` | When the role declares Compose volumes | socialhome (new), magento. |
| [x] | `meta/schema.yml` | When the role declares credentials | socialhome (new); magento's deleted — it declared two Adobe keys with `^.*$` validation that passed empty values silently. |
| [x] | `tasks/main.yml` | Usually | socialhome (new), listmonk, magento. |
| [x] | `templates/compose.yml.j2` | For containerized app roles | socialhome (new), magento. |
| [x] | `templates/env.j2` | When the app uses environment files | socialhome (new). |
| [ ] | `templates/style.css.j2` | When the role injects branding | Not applicable. |
| [ ] | `templates/javascript.js.j2` | When the role injects frontend behavior | Deliberately omitted for socialhome — see Change Details. listmonk ships `files/oidc-autologin.js` through a settings key instead. |
| [x] | `meta/users.yml` | When the role bootstraps users | socialhome (new). |
| [ ] | `files/Dockerfile` | When a custom image is required | Not applicable. |
| [x] | `templates/playwright.env.j2` | When Playwright coverage is included | **The centrepiece.** 21 new, 46 amended. |
| [x] | `files/playwright/playwright.spec.js` | When Playwright coverage is included | socialhome's nav labels corrected; mastodon's companion guarded. |

### Registered

| Check | Item | Purpose |
|---|---|---|
| [x] | Per-entity ports | socialhome `local.http: 8089` / `internal.http: 8099`, verified unique repo-wide. |
| [x] | Per-role subnet | socialhome `192.168.74.0/24`, verified non-overlapping against all 100 role networks. |

---

## Local Validation

* [x] Deployment target and distro documented
* [ ] Playwright test run documented
* [ ] Login flow tested
* [ ] Logout flow tested
* [ ] Screenshot attached when the change is user-visible

**Read this section carefully — no deploy was run.** Three unchecked boxes are the honest state of this PR.

What *was* executed:

* `tests/lint` + `tests/integration`: **532 passed, 1548 subtests**; `test-unit`: **4036 passed, 477 subtests** (host runner).
* `make lint-playwright`: **82/82 specs** staged and registered exactly as the deploy runner does.
* 256 Jinja expressions in the new templates diffed against the 61 already rendering in CI: 11 unprecedented, all one already-executed shape; **zero unresolvable lookup paths**.
* `listmonk/listmonk:v6.2.0` booted against a throwaway Postgres and probed: `GET /` → 200 HTML with a Login link, `HEAD /` → 404, and the OIDC form appears only after a reload. The full chain driven by `clickOidcLoginLink`'s own source in a real browser: no reload → no form; reload alone → not reached; reload + `custom_js` → **reached**.
* `ghcr.io/social-home-io/socialhome:2026.6.16` pulled and probed: `http_status` argv **exit 0** inside the container, `/healthz` returns byte-for-byte the shape the spec asserts, `POST /api/setup/standalone` → 201 then 409, `setup_required` flips true → false, `[standalone].external_url` lands in the pairing `inbox_url`, container runs as uid 10001.

The containerized `make test` could not run here: three dev stacks cannot share one Docker address pool. The host runner covers the same suites.

---

## Security Impact

* [x] Security impact present

* **Affected surfaces:** Listmonk's OIDC becomes active for the first time (it was silently disabled), and its S3 credentials reach the bucket for the first time. socialhome adds a publicly reachable vhost with no SSO gate — required, because oauth2-proxy would break federation. socialhome's CSP whitelists `blob:` on `script-src` for an audio worklet.
* **Risk reduction / new exposure:** Net reduction on listmonk — an instance that believed it had SSO had none. New exposure on socialhome: an unauthenticated federation endpoint, which is the protocol's design. Without `web-svc-coturn` the app falls back to Google's STUN server, which the README now states.
* **Security-specific validation:** the CSP header was rendered and read but **never enforced by a browser**. hCaptcha on listmonk turned out never to have been active (three settings keys do not exist in v6.2.0); this PR does not switch it on — that is recorded in the role's `TODO.md` as a deliberate deferral.

---

## Review Focus

* **Highest-risk:** `roles/web-app-socialhome/` in full — nothing in it has been rendered by Ansible or deployed. It enters the CI matrix at `alpha` having never run anywhere, and the first CI round is its first execution. `modes.swarm.enabled: false` is deliberate: the orchestrator chains swarm → compose → host under `mode_fail_fast`, so a red swarm job there would skip 92 unrelated jobs. Flip it once compose is green.
* **Second:** the 17 newly-armed specs. Static analysis says several will be red on first contact. That is the point of arming them, but it will make the first run noisy.
* **Migration / rollback:** listmonk's settings task now recreates the container on every deploy (`changed_when: true`), including when nothing changed. Costs a restart, breaks nothing. Rollback of any single role is its own commit.
* **Specific feedback requested:**
  1. `roles/web-app-listmonk/meta/server.yml` `status_codes.default: 404` is **correct** and was twice misjudged during this work. `GET /` is 200 HTML; the health probe uses `requests.head`, and Echo does not synthesise HEAD from GET, so `HEAD /` falls into `RouteNotFound`. Please do not "fix" it to 200 — that breaks the post-deploy gate for `newsletter.*`.
  2. Nine of the 89 rationales state plainly that the block is **not** a role property but a gap in the shared persona helpers (decidim, fider, jenkins, taiga, wordpress, xwiki, odoo, penpot, opentalk). Those are recorded coverage debt, and a design for closing them exists.
  3. `web-app-bookwyrm` provisions **no staff account at all** — found while writing its rationale. Its block is correct today; the repair is bootstrapping a user, not fixing a selector.

---

## Definition of Done (DoD)

---

## Additional Notes

Every finding in this PR came out of an adversarial review loop, and several of the fixes were themselves refuted before landing. Three examples worth carrying into review, because they are the ones most likely to be re-broken:

* The first version of the env-template lint had an opt-out the docstring told you to write as `# nocheck:` on line 1 of a `.js` file — a hard parse error — and a substring check that any prose mentioning the rule name disarmed. Both were found by executing them, not reading them.
* Listmonk's `/` status was "corrected" twice in opposite directions before the container was actually booted and probed.
* `worker-src: blob:` in socialhome's CSP protected nothing; the bundle registers no service worker. The real need is `script-src`, for an audio worklet.

`security.enable_captcha`, `security.captcha_key` and `security.captcha_secret` in listmonk's settings do not exist in v6.2.0 — the live table carries one `security.captcha` object — so all three `UPDATE`s match zero rows. Repairing that would switch a captcha on for the public subscription form, so it is deferred to the role's `TODO.md` rather than smuggled in here.


kevinveenbirkenbach:
Merged to https://github.com/infinito-nexus/core/pull/380 to don't work on to much things parallel and use CI efficient.

PR #611, opened 2026-09-10
feat(web-app-openbao): integrate OpenBao as a first-class role

evangelostsak:
## Summary

Adds `web-app-openbao`, a new `web-app-*` role that runs [OpenBao](https://openbao.org/) as a first-class Infinito.Nexus application: a self-hosted secrets manager reachable at its own vhost, with Keycloak OIDC and LDAP sign-in, three RBAC policy tiers, a KV v2 store, an optional internal PKI, and Prometheus native-metrics coverage. It deploys in both `compose` and `swarm` mode.

The branch also carries one shared-infrastructure fix that the swarm variant depends on (see *Change Details*).

Implements [OpenProject #628](https://project.infinito.nexus/work_packages/628) — *Integrate OpenBao into Infinito.Nexus*.
Requirement: [`docs/requirements/032-web-app-openbao.md`](https://github.com/evangelostsak/infinito.nexus/blob/feature/web-app-openbao/docs/requirements/032-web-app-openbao.md) — 60 of 64 acceptance criteria checked; the four open ones are listed under *Additional Notes*.

---

## Template Type

Select the primary intent of this PR:

* [x] **Feature** - Adds or extends server functionality
* [ ] **Fix** - Repairs broken or incorrect server behavior

---

List the impacted roles and related services.

* Primary `web-*` role(s): `web-app-openbao` (new)
* Related `web-svc-*`, `sys-front-*`, `svc-db-*`, auth, mail, proxy, or storage role(s):
  * `web-app-keycloak` — OIDC client, RBAC group paths, `run_after` dependency
  * `svc-db-openldap` — LDAP auth method, `run_after` dependency
  * `web-app-prometheus` — native-metrics scrape job plus a new `OpenBaoSealed` alert rule in the shared alert template
  * `svc-prx-openresty` / `sys-stk-front-proxy` — vhost, and a `deny all` on `/v1/sys/metrics`
  * `sys-svc-compose` — network pre-creation fix required by the swarm variant
  * `web-svc-logout`, `web-svc-css`, `web-app-matomo`, `web-app-dashboard` — standard integrations
  * `svc-net-tor` — onion vhost variant

## Preferred Integrations

Integrate the change into the following services when possible:

* [x] Dashboard
* [x] Matomo
* [x] OIDC
* [x] LDAP
* [x] Logout

---

```yaml
roles:
  # - web-app-nextcloud
  # - web-app-matomo
  # - sys-version
```

---

## Change Type

Select the semantic version impact of this change:

---

### What problem does this solve?

There was no self-hosted secrets manager in the platform. Secrets lived only in the encrypted inventory, with no runtime API, no per-role read scoping, and no place to issue internal certificates from. `web-app-openbao` provides that surface, gated by the same Keycloak identity every other role uses.

### Upstream image

`openbao/openbao:2.6.2`, pinned to the immutable tag.

### How the role works

* **Storage** — raft integrated storage on a named volume mounted at `/openbao/file`. That path is deliberate: the image only chowns `/openbao/{config,logs,file}`, so the more obvious `/openbao/data` stays root-owned and the server cannot write to it. Logs get their own named volume so no anonymous volume is created.
* **Sealing** — `seal "static"` with the key injected as `env://OPENBAO_SEAL_KEY` from the encrypted inventory, so a restarted container comes back unsealed without operator interaction. Recovery keys generated at `bao operator init` are discarded on purpose.
* **Bootstrap** — first deploy initialises the node, then creates an `ansible` AppRole with a caller-pinned `role_id` / `custom-secret-id` and revokes the root token. Every later deploy authenticates through that AppRole, so **no root token is ever stored**.
* **Auth** — `jwt`/`oidc` against the Keycloak realm plus `ldap` against the directory. Keycloak group paths and LDAP groups are bound to OpenBao external identity groups, which carry the policies.
* **RBAC** — three policy tiers rendered as templates: `administrator`, `operator`, `reader`. A user in none of the groups receives only `default` and reaches no privileged path.
* **PKI** — optional (`services.openbao.pki.enabled`, off by default, exercised by variant 2): root CA, issuing URLs, and an `internal` role.
* **Login/logout** — login is the Keycloak OIDC journey through the OpenBao UI, or LDAP username/password. Logout integrates with the `logout` service (proxy activation, universal-logout injection with its CSP hash, `LOGOUT_DOMAINS` membership). See *Additional Notes* for a session-termination finding.
* **Proxying** — the listener serves `/v1/sys/metrics` unauthenticated so Prometheus can scrape it over the container network; the front proxy `deny all`s that exact location so it is not reachable publicly. Asserted by `test-rbac-denial.js`.
* **Swarm** — `placement: manager` (raft's BoltDB store is mmap + fsync based, so it cannot live on NFS and must stay with its volume), `service_update_order = 'stop-first'`, single replica.

### The `sys-svc-compose` fix

`web-app-prometheus` pre-creates the docker network of every native-metrics app so it can attach as a scraper. `roles/sys-svc-compose/tasks/utils/network/create.yml` stamped only `com.docker.compose.network` and `com.docker.compose.project`. Those labels make `docker compose up` adopt an existing network, but `docker stack deploy` looks for a different one: it lists candidate networks filtered on `com.docker.stack.namespace=<stack>`, sees nothing, and falls through to creating the network again:

```
failed to create network openbao: Error response from daemon: network with name openbao already exists
```

The swarm branch now also stamps `com.docker.stack.namespace=<entity>`. This is a latent bug in shared infrastructure rather than an OpenBao-specific one — any role whose network is pre-created before its own stack deploys hits it; OpenBao was simply the first native-metrics app ordered after Prometheus in swarm.

### Alternatives considered

* **Shamir unsealing instead of a static seal** — rejected: it needs a human to unseal after every restart, which no unattended deploy can satisfy.
* **Keeping the root token for later deploys** — rejected: a long-lived root token in the inventory is a worse blast radius than an AppRole with a 30 m token TTL.
* **`/openbao/data` as the raft path** — does not work, see above.
* **Alerting on `vault_core_unsealed`** — does not work, see *Additional Notes*.

---

## File Checklist

| Check | Item | When to include | Purpose |
|---|---|---|---|
| [x] | `README.md` | Usually | Documents role-specific usage, setup notes, and contributor context. |
| [x] | `meta/main.yml` | Usually | Declares Ansible Galaxy info and `dependencies:`. No project-internal `run_after:` / `lifecycle:` (those live on the primary entity in `meta/services.yml`). |
| [x] | `vars/main.yml` | Usually | Defines the shared fixed role variables as the main source of truth. |
| [x] | `meta/services.yml` | Usually | Per-entity service config (file root IS the services map keyed by `<entity_name>`). Holds image/version, `ports.{internal,local,public}`, `run_after`/`lifecycle` on the primary entity, plus inlined per-service settings. |
| [x] | `meta/server.yml` | Usually | Server-level config (file root IS `applications.<app>.server`). CSP, `domains`, `status_codes`, plus the per-role `networks.local.{subnet,dns_resolver}`. |
| [x] | `meta/rbac.yml` | When the role declares RBAC | RBAC declarations (file root IS `applications.<app>.rbac`). |
| [x] | `meta/volumes.yml` | When the role declares Compose volumes | Volumes map (file root IS the volumes map; no `compose:`/`volumes:` wrapper). |
| [x] | `meta/secrets.yml` | When the role declares credentials | Credential schema with optional `default:` field. |
| [x] | `tasks/main.yml` | Usually | Acts as the role entry point and includes the main task flow. |
| [x] | `templates/compose.yml.j2` | For containerized app roles | Defines the service, volume, environment, port, and network wiring. |
| [x] | `templates/env.j2` | When the app uses environment files | Renders the app environment configuration. |
| [ ] | `templates/style.css.j2` or `files/style.css` | When the role injects custom branding or theming | Not included — the role uses the shared `web-svc-css` injection. |
| [ ] | `templates/javascript.js.j2` or `files/javascript.js` | When the role injects custom frontend behavior | Not included — no role-local frontend behaviour is needed. |
| [x] | `meta/users.yml` | When the role bootstraps users or identities | Role-local user definitions (file root IS the users map; no `users:` wrapper). |
| [ ] | `files/Dockerfile` | When a custom image is required | Not included — the upstream image is used unmodified. |
| [x] | `templates/playwright.env.j2` | When Playwright coverage is included | Configures the Playwright test environment. |
| [x] | `files/playwright/playwright.spec.js` | When Playwright coverage is included | Defines the Playwright login and logout test flow. |

### Registered

| Check | Item | When to include | Purpose |
|---|---|---|---|
| [x] | Per-entity ports in `meta/services.yml.<entity>.ports.{internal,local,public}` | When the app exposes a service | `internal.http: 8200`, `local.http: 8081`. |
| [x] | Per-role subnet in `meta/server.yml.networks.local.subnet` | When the app communicates over container networks | `192.168.221.0/24`, declared in `meta/networks.yml`. |

---

## Local Validation

Describe how the change was validated locally.

* [x] Deployment target and distro documented
* [x] Playwright test run documented
* [x] Login flow tested
* [x] Logout flow tested
* [ ] Screenshot attached when the change is user-visible

**Compose** — deployed repeatedly on `debian`; the role-local Playwright suite runs green (14 passed in fork CI, see below). Idempotency was measured down to 0 changed tasks on a second run.

**Swarm** — validated on a live `swarm-zombie` cluster (`variant=0`, `debian`, 1 manager + 2 workers + NFS + backup node). After the `sys-svc-compose` fix:

* the `openbao` overlay carries `{"com.docker.compose.network":"default","com.docker.compose.project":"openbao","com.docker.stack.namespace":"openbao"}`
* `docker service ls` → `openbao_openbao replicated 1/1`
* `bao status` → `Initialized true`, `Sealed false`, `HA Mode active`, `Storage Type raft`
* the whole role reconciled: KV mount, all three RBAC policies, OIDC and LDAP auth methods, identity groups and both alias sets, PKI, token cleanup
* the app answers `307` (into its UI) through the node-local proxy

The fix itself was also verified in isolation against docker 29.1.3 on a throwaway swarm: with only the compose labels the deploy reproduces the error above; with the namespace label the stack adopts the *same network id* rather than recreating it; and an external-network consumer stack deploying with `--prune` leaves that network intact, so shared provider overlays do not regress.

`tests/lint`: 356 passed, 1 skipped, 326 subtests.

---

## Security Impact

Indicate whether this change has security implications.

* [x] Security impact present

**Affected auth, TLS, permissions, secrets, headers, or exposed surfaces:**

* Introduces a secrets manager as a public vhost — the highest-value surface in the deployment.
* Three new inventory credentials (`seal_key`, `approle_role_id`, `approle_secret_id`). **The seal key is the whole ballgame**: every backup of the raft volume is undecryptable without it, and its authoritative home is the encrypted inventory.
* `/v1/sys/metrics` is served without authentication on the listener so the internal scrape works, and is `deny all`ed at the front proxy so it is not publicly readable.
* Keycloak group paths and LDAP groups are mapped to OpenBao identity groups, so RBAC tiers follow the existing directory.

**Risk reduction, new exposure, or compatibility considerations:**

* No root token is persisted anywhere after bootstrap; the reconciliation path is an AppRole with `token_ttl=30m token_max_ttl=1h`, and the stored token is dropped from the container at the end of every run.
* Recovery keys from `bao operator init` are deliberately discarded, which means the AppRole is the sole administrative path. That is a conscious trade: it removes a set of long-lived keys from the deployment at the cost of making a lost seal key unrecoverable. Documented in the role README.
* The `sys-svc-compose` change adds a label to pre-created networks. It does not widen network reachability; it only lets the owning stack recognise a network it would otherwise refuse to adopt.

**Security-specific validation performed:**

* `test-rbac-denial.js` asserts that unauthenticated calls to privileged paths are refused, that an invalid token is refused, that the three policies exist and are distinct, and that `/v1/sys/metrics` is not served through the public proxy (verified live: **403** on the onion vhost).
* `test-guest-persona.js` asserts a guest never reaches an authenticated surface; `test-biber-persona` is blocked by design, since a user holding no OpenBao RBAC group gets only the `default` policy.
* `test-ldap-login.js` asserts that a user in no OpenBao role group receives no privileged policy.

---

## Review Focus

* **Highest-risk files, roles, or flows:**
  * `roles/sys-svc-compose/tasks/utils/network/create.yml` — the only file outside the new role that changes behaviour for *every* role. Worth a second opinion on the label value (`network_role_id | get_entity_name`, i.e. the owning stack's namespace, not the deploying stack's).
  * `roles/web-app-openbao/tasks/01_init.yml` and `02_auth.yml` — the bootstrap-once path, root-token revocation, and AppRole pinning.
  * `roles/web-app-openbao/templates/openbao.hcl.j2` — `disable_mlock = true` (raft and mlock are mutually exclusive upstream) and `cluster_addr` on loopback (swarm renames every task, and a single-replica raft node only advertises to itself).
* **Migration, rollback, or security-sensitive concerns:** the seal key must reach the inventory before the first deploy and must never be rotated without re-keying; backup/restore of the raft volume is documented in the role README.
* **Specific feedback requested from reviewers:** whether discarding the recovery keys is the right default, or whether they should be stored alongside the seal key.

---

## Definition of Done (DoD)

* [x] The implementation follows the Definition of Done, and the contribution guidelines in [CONTRIBUTING.md](https://github.com/infinito-nexus/core/blob/main/CONTRIBUTING.md) were considered and applied during implementation.

---

## Additional Notes

**Draft, on purpose.** The last fork CI run (`34327141502`) was red: the two compose variants passed with the OpenBao suite at 14 passed, and the swarm variant failed on exactly the network collision this branch now fixes. No fork CI run has gone green since the fix, so this stays a draft until one does.

Four acceptance criteria in the requirement document remain open, each with its blocker recorded:

1. `make swarm-playwright role=web-app-openbao` has never completed a full run. The swarm *deploy* is now verified (above), but the spec suite needs one uninterrupted swarm run — the local drill is killed by the step budget (`rc=137`) because every co-deployed role's onion spec burns repeated 300 s timeouts on the dev machine first.
2. **Logout terminates the session** is only half proven. The integration half is verified. The termination half is not: navigating to `/ui/vault/logout` leaves the session intact and the UI returns to `/ui/vault/secrets`, because the Ember app restores its token from browser storage. Proving termination needs the UI's own logout control driven from an authenticated session. The assertion was reverted rather than left failing, and the finding is recorded in the role README.
3. **A sealed instance fires an alert.** An `OpenBaoSealed` rule now exists in the shared Prometheus alert template, but the firing condition was never observed — that needs an instance deliberately sealed, which the AppRole-only admin model makes awkward to arrange. Note for reviewers: OpenBao keeps Vault's metric namespace, and `vault_core_unsealed` is **unusable** as the alert source — a healthy, demonstrably unsealed node emits it as `vault_core_unsealed{cluster=""} 0`, so `== 0` would fire permanently. The rule uses `vault_core_active` with `/v1/sys/health` as the primary signal.
4. **Cross-link to this PR** — closed by opening it; the requirement document is updated with this PR's URL.


roles/sys-svc-compose/tasks/utils/network/create.yml
```diff
@@ -39,6 +39,10 @@
       - "com.docker.compose.network={{ lookup('shared_network_key', network_role_id) }}"
       - --label
       - "com.docker.compose.project={{ network_role_id | get_entity_name }}"
```
kevinveenbirkenbach, line 41: no narrative exception tag. Delete comment :)

roles/web-app-openbao/files/playwright/test-ldap-login.js
```diff
@@ -0,0 +1,54 @@
...
+
+test("ldap: a user in no OpenBao role group gets no privileged policy", async ({ request }) => {
+  skipUnlessServiceEnabled("ldap");
+
+  const response = await request.post(
+    `${baseUrl}/v1/auth/ldap/login/${encodeURIComponent(biberUsername)}`,
+    { data: { password: biberPassword }, failOnStatusCode: false, timeout: resolveTimeout(30_000) },
+  );
+
+  // Exception: rejection and admission-with-`default` both satisfy the contract; only a
```
kevinveenbirkenbach: no exception

roles/web-app-openbao/files/playwright/test-oidc-login.js
```diff
@@ -0,0 +1,124 @@
...
+const adminUsername = decodeDotenvQuotedValue(process.env.ADMIN_USERNAME || "");
+const adminPassword = decodeDotenvQuotedValue(process.env.ADMIN_PASSWORD || "");
+const issuerUrl = normalizeBaseUrl(process.env.OIDC_ISSUER_URL || "");
+
+const AUTH_PATH = "/ui/vault/auth?with=oidc";
+const CALLBACK_PATH = "/ui/vault/auth/oidc/oidc/callback";
+
+test.use({ ignoreHTTPSErrors: true });
+
+// Exception: a failed auth_url request renders as "Invalid role", indistinguishable from a
```
kevinveenbirkenbach: no exception

roles/web-app-openbao/files/playwright/test-oidc-login.js
```diff
@@ -0,0 +1,124 @@
...
+  ).toContain("openid-connect/auth");
+});
+
+test("oidc: the administrator signs in through Keycloak and lands in the OpenBao UI", async ({ page }) => {
+  skipUnlessServiceEnabled("sso");
+  // Exception: the default per-test budget is smaller than this flow's own onion-scaled waits.
+  test.setTimeout(resolveTimeout(240_000));
+  await page.context().clearCookies();
+
+  // Exception: the provider is a second hidden service needing its own Tor circuit; building it
```
kevinveenbirkenbach: no exception

roles/web-app-openbao/files/playwright/test-oidc-login.js
```diff
@@ -0,0 +1,124 @@
...
+          }
+        }, CALLBACK_PATH),
+      {
+        timeout: resolveTimeout(60_000),
+        message: "the auth_url endpoint the OIDC form depends on never became reachable",
+      },
+    )
+    .toBe(200);
+
+  // Exception: clicking before the component's own debounced role fetch settles opens no popup
```
kevinveenbirkenbach: no exception

roles/web-app-openbao/files/playwright/test-rbac-denial.js
```diff
@@ -0,0 +1,53 @@
...
+    failOnStatusCode: false,
+    timeout: resolveTimeout(30_000),
+  });
+  expect(
+    [400, 403],
+    `expected an invalid token to be refused, got ${response.status()}`,
+  ).toContain(response.status());
+});
+
+// Exception: the listener serves metrics unauthenticated for the internal scrape, so the
```
kevinveenbirkenbach: no exception

roles/web-app-openbao/meta/rbac.yml
kevinveenbirkenbach, line 1: Are there tests for all of this groups? Othewrwise please add. 
You can add the LAM role in the variants.yml (just variant not services.yml) and add, remove the biber user to this group by logging in as administrator to lam via playwright and verify afterwards that the user does the actions which you assume they can do.

roles/web-app-openbao/meta/server.yml
```diff
@@ -0,0 +1,8 @@
+---
+# Exception: OpenBao answers `/` with a 307 into its UI, which the health checker does not expect by default.
```
kevinveenbirkenbach: no exception

roles/web-app-openbao/meta/services.yml
```diff
@@ -0,0 +1,88 @@
...
+  shared: "{{ 'web-app-dashboard' in group_names }}"
+matomo:
+  bond: 1
+  enabled: "{{ 'web-app-matomo' in group_names }}"
+  shared: "{{ 'web-app-matomo' in group_names }}"
+css:
+  bond: 1
+  enabled: "{{ 'web-svc-css' in group_names }}"
+  shared: "{{ 'web-svc-css' in group_names }}"
+# nocheck: email, dynamic-flag - OpenBao has no outbound mail surface: it notifies nothing and owns no user mailboxes; audit trails leave through the logging stack.
```
kevinveenbirkenbach, line 28: sure that it doesn't send system emails? If it doesn't send system emails this is correct, if it does send system emails it would be required to be activated.

roles/web-app-openbao/tasks/utils/group_alias.yml
```diff
@@ -0,0 +1,25 @@
+---
+# Exception: a group alias is keyed by (name, mount_accessor) and creating a duplicate errors,
```
kevinveenbirkenbach: No Exception

roles/web-app-openbao/tasks/00_core.yml
```diff
@@ -0,0 +1,45 @@
+---
+# Exception: the config templates notify `compose-up`, so they MUST be rendered after
```
kevinveenbirkenbach: no exception

roles/web-app-openbao/tasks/00_core.yml
```diff
@@ -0,0 +1,45 @@
...
+        dest: "{{ [OPENBAO_POLICY_DIR_HOST, 'administrator.hcl'] | path_join }}"
+        mode: "0644"
+      - src: "policies/operator.hcl.j2"
+        dest: "{{ [OPENBAO_POLICY_DIR_HOST, 'operator.hcl'] | path_join }}"
+        mode: "0644"
+      - src: "policies/reader.hcl.j2"
+        dest: "{{ [OPENBAO_POLICY_DIR_HOST, 'reader.hcl'] | path_join }}"
+        mode: "0644"
+
+- name: "🚿 flush handlers for '{{ application_id }}'"
```
kevinveenbirkenbach, line 31: could make sense to move this after the flush of the next task to prevent one extra flush. Not 100% sure.

roles/web-app-openbao/tasks/01_init.yml
```diff
@@ -0,0 +1,109 @@
...
+
+- name: "⏳ Wait until the OpenBao API answers"
+  ansible.builtin.shell: > # nocheck: container-exec-resolver  runtime-resolved CID, delegated to the hosting node
+    container exec {{ OPENBAO_LOCAL_CID | quote }}
+    bao status -format=json
+  register: openbao_status_raw
+  changed_when: false
+  retries: 30
+  delay: 5
+  # Exception: exit 2 means sealed and exit 1 means not yet initialised; both prove the API answers.
```
kevinveenbirkenbach: No Exception

roles/web-app-openbao/tasks/01_init.yml
```diff
@@ -0,0 +1,109 @@
...
+    file: 05_pki.yml
+    apply:
+      delegate_to: "{{ OPENBAO_HOST_NODE }}"
+  when: OPENBAO_PKI_ENABLED | bool
+
+- name: "🧹 Drop the stored reconciliation token from the container"
+  ansible.builtin.shell: > # nocheck: container-exec-resolver  runtime-resolved CID, delegated to the hosting node
+    container exec {{ OPENBAO_LOCAL_CID | quote }}
+    sh -c 'bao token revoke -self >/dev/null 2>&1; rm -f ~/.bao-token'
+  # Exception: teardown of this run's own short-lived token, not configuration drift.
```
kevinveenbirkenbach: No Exception

roles/web-app-openbao/tasks/03_policies.yml
```diff
@@ -0,0 +1,32 @@
+---
+- name: "🗄 Ensure the KV v2 secrets engine is mounted at '{{ OPENBAO_KV_MOUNT }}'"
+  ansible.builtin.shell: > # nocheck: container-exec-resolver  runtime-resolved CID, delegated to the hosting node
+    container exec {{ OPENBAO_LOCAL_CID | quote }}
+    sh -c 'bao secrets list -format=json | grep -q "\"{{ OPENBAO_KV_MOUNT }}/\"" ||
+    { bao secrets enable -path={{ OPENBAO_KV_MOUNT | quote }} -version=2 kv && echo OPENBAO_CHANGED; }'
+  register: openbao_kv_mount
+  changed_when: "'OPENBAO_CHANGED' in openbao_kv_mount.stdout"
+
+# Exception: `bao policy read` reformats the HCL, so the rendered file cannot be diffed against
```
kevinveenbirkenbach: No Exception

roles/web-app-openbao/tasks/04_auth_methods.yml
```diff
@@ -0,0 +1,143 @@
...
+        container exec -i {{ OPENBAO_LOCAL_CID | quote }}
+        bao write auth/oidc/config
+        oidc_discovery_url={{ OPENBAO_OIDC_ISSUER | quote }}
+        oidc_client_id={{ OPENBAO_OIDC_CLIENT_ID | quote }}
+        oidc_client_secret=-
+        default_role={{ OPENBAO_OIDC_DEFAULT_ROLE | quote }}
+      args:
+        stdin: "{{ OPENBAO_OIDC_CLIENT_SECRET }}"
+        stdin_add_newline: false
+      # Exception: the write always runs so a rotated client secret lands, but OpenBao redacts
```
kevinveenbirkenbach: No Exception

roles/web-app-openbao/tasks/04_auth_methods.yml
```diff
@@ -0,0 +1,143 @@
...
+        groupdn={{ LDAP.DN.OU.ROLES | quote }}
+        groupattr=cn
+        binddn={{ LDAP.DN.ADMINISTRATOR.DATA | quote }}
+        bindpass=-
+        insecure_tls=false
+        starttls=false
+      args:
+        stdin: "{{ LDAP.BIND_CREDENTIAL }}"
+        stdin_add_newline: false
+      # Exception: as with the OIDC config, the bind password is write-only, so the write always
```
kevinveenbirkenbach: No Exception

roles/web-app-openbao/tasks/05_pki.yml
kevinveenbirkenbach, line 1: Just fopr clearification:
Why is this necessary?
Will this also work in production with e.g. letsencrypt?

roles/web-app-openbao/templates/policies/administrator.hcl.j2
```diff
@@ -0,0 +1,48 @@
...
+
+path "sys/health" {
+  capabilities = ["read"]
+}
+
+path "sys/seal-status" {
+  capabilities = ["read"]
+}
+
+# Exception: without these the web UI renders an empty secrets list.
```
kevinveenbirkenbach: No Exception

roles/web-app-openbao/templates/policies/operator.hcl.j2
```diff
@@ -0,0 +1,32 @@
...
+
+path "sys/health" {
+  capabilities = ["read"]
+}
+
+path "sys/seal-status" {
+  capabilities = ["read"]
+}
+
+# Exception: without these the web UI renders an empty secrets list.
```
kevinveenbirkenbach: No Exception

roles/web-app-openbao/templates/policies/reader.hcl.j2
```diff
@@ -0,0 +1,24 @@
...
+
+path "sys/health" {
+  capabilities = ["read"]
+}
+
+path "sys/seal-status" {
+  capabilities = ["read"]
+}
+
+# Exception: without these the web UI renders an empty secrets list.
```
kevinveenbirkenbach: No Exception

roles/web-app-openbao/templates/openbao.hcl.j2
```diff
@@ -0,0 +1,35 @@
...
+}
+
+telemetry {
+  prometheus_retention_time = "24h"
+  disable_hostname          = true
+}
+
+api_addr = "{{ OPENBAO_BASE_URL }}"
+
+# Exception: loopback, not the container name — swarm renames every task, and a single-replica raft node only advertises to itself.
```
kevinveenbirkenbach: No Exception

roles/web-app-openbao/templates/openbao.hcl.j2
```diff
@@ -0,0 +1,35 @@
+ui = true
+
+log_level = "{{ OPENBAO_LOG_LEVEL }}"
+
+# Exception: raft and mlock are mutually exclusive upstream.
```
kevinveenbirkenbach: No Exception

roles/web-app-prometheus/templates/configuration/alert_rules.yml.j2
```diff
@@ -64,6 +64,26 @@ groups:
             This is a communication-critical application.
 {% endif %}
 
+{% if 'web-app-openbao' in lookup('native_metrics_apps') %}
+  - name: openbao_seal
+    interval: 30s
+    rules:
+      # Exception: `vault_core_unsealed` is unusable here — an unsealed node emits it as
```
kevinveenbirkenbach: No Exception

roles/web-app-prometheus/templates/configuration/alert_rules.yml.j2
kevinveenbirkenbach, line 1: isn't there a convention existing which allows you to place this in the role? goes a bit against SPOT, SRP, atomic to keep it here. 

---> checked it out seems like no convention exist. 

Propably idea would be to implement one. e.g.

loops over all in group_name roles and places their templates/prometheus/alert_rules.yml.j2 in here.

That allows to keep it scalable for the future and allows other apps as well to add their own additional alert rules.

kevinveenbirkenbach (changes requested):
Looks over all really good. 

Still required:
- SPOT for prometheus alerts
- Playwright tests for all rbac groups. 

Please activate in your github repository the saving of all video records (by variable) and add them for all the playwright tests of your new role. Just add them to the openproject item and link them here)  @AlejandroRomanIbanez created a fix because it seem's like the videos hadn't been saved. 

Cherry pick it from him so that it works. 

As soon as the CI finished verify that the jobs which fail in your CI fail due to the same reason like in https:/s.infinito.nexus/integrationstatus . Add issues for the failing jobs or comment in the existing ones. 

If they fail due to different reasons you MUST solve them :) 

Thanks a lot :)

roles/web-app-lam/meta/csp.yml
kevinveenbirkenbach, line 1: Deeper research why this is just necessary now.

roles/web-app-openbao/meta/services.yml
```diff
@@ -0,0 +1,87 @@
...
+prometheus:
+  bond: 1
+  enabled: "{{ 'web-app-prometheus' in group_names }}"
+  shared: "{{ 'web-app-prometheus' in group_names }}"
+  native_metrics:
+    enabled: true
+    port: 8200
+openbao:
+  # nocheck: default-placement-manager  raft keeps its BoltDB store mmap'd with fsync durability, so it cannot live on NFS; the node is pinned in swarm to stay with its volume
+  placement: manager
```
kevinveenbirkenbach, line 42: Manager is probably wrong here because then it'Äs just placed on the manager node. 
If it's related to volume then make to volume swarm compatible.

roles/web-app-openbao/meta/services.yml
```diff
@@ -0,0 +1,87 @@
...
+  mem_reservation: 128m
+  mem_limit: 512m
+  pids_limit: 512
+  min_storage: 60MB
+  image: openbao/openbao
+  version: "2.6.2"
+  backup:
+    no_stop_required: false
+  # nocheck: single-replica  single-node raft: a second replica would need a real raft join/bootstrap flow, and two tasks against one BoltDB directory corrupt the store
+  replicas: 1
```
kevinveenbirkenbach, line 58: make bootstrap flow just running on one host eg IS_STACK_HOST guarded.

roles/web-app-openbao/meta/services.yml
```diff
@@ -0,0 +1,87 @@
...
+      http: 8200
+    local:
+      http: 8093
+  healthcheck:
+    flavor: wget
+    path: "v1/sys/health?uninitcode=200&sealedcode=200&standbyok=true"
+  run_after:
+    - web-app-keycloak
+    - svc-db-openldap
+  lifecycle: alpha
```
kevinveenbirkenbach: beta

kevinveenbirkenbach (commented):
Mainly placement issue

roles/web-app-openbao/meta/volumes.yml
kevinveenbirkenbach, line 1: Concerning swarm placement via replica 1 or manager:
Chance is hiugh that you need to place a volume here to make it swarm capable.

roles/web-app-openbao/tasks/02_auth.yml
```diff
@@ -0,0 +1,69 @@
...
+    bao write auth/approle/role/{{ OPENBAO_ANSIBLE_APPROLE | quote }}/role-id
+    role_id={{ OPENBAO_APPROLE_ROLE_ID | quote }}
+  changed_when: true
+  delegate_to: "{{ OPENBAO_HOST_NODE }}"
+
+- name: "🔑 Pin the SecretID to the inventory value"
+  ansible.builtin.shell: > # nocheck: container-exec-resolver  runtime-resolved CID, delegated to the hosting node
+    container exec -i {{ OPENBAO_LOCAL_CID | quote }}
+    bao write auth/approle/role/{{ OPENBAO_ANSIBLE_APPROLE | quote }}/custom-secret-id
+    secret_id=-
```
kevinveenbirkenbach, line 52: Verify that - reads from stdin.

roles/web-app-openbao/templates/prometheus/alert_rules.yml.j2
```diff
@@ -0,0 +1,16 @@
+{% if 'web-app-openbao' in lookup('native_metrics_apps') %}
```
kevinveenbirkenbach: By feeling: 
MAy this is redundant: The prometheus loop should already filter it out and don't integrate alerts for apps which aren't deployed and don't have it integrated.

roles/web-app-openbao/README.md
```diff
@@ -0,0 +1,146 @@
...
+      --inventory-file "$INVENTORY/devices.yml" \
+      --host "$HOST" \
+      --include "$APP" \
+      --vars "{\"TLS_MODE\": \"$TLS_MODE\", \"DOMAIN_PRIMARY\": \"$DOMAIN\", \"users\": {\"administrator\": {\"authorized_keys\": [\"$SSH_PUBLIC_KEY\"]}}}" &&
+    infinito administration deploy dedicated "$INVENTORY/devices.yml" \
+      --password-file "$INVENTORY/.password" \
+      --diff -vv'
+```
+
+## Developer Notes
```
kevinveenbirkenbach, line 107: To much informatzion. Nobody will reads this :)

roles/web-app-prometheus/templates/configuration/alert_rules.yml.j2
```diff
@@ -76,3 +76,8 @@ groups:
           description: >
             Prometheus target {{ '{{' }} $labels.instance {{ '}}' }} (job: {{ '{{' }} $labels.job {{ '}}' }})
             has been unreachable for more than 5 minutes.
+
+{% for alert_rules_role_id in lookup('alert_rule_roles') %}
```
kevinveenbirkenbach, line 80: filter just by activated

kevinveenbirkenbach (commented):
Seems overall good. 

Alert loop MUST be optimized and from previous review full swarm capability. 

Out of Scope for this PR: 
Integrate into all invokable roles. See https://open.project.infinito.nexus/projects/infinito/work_packages/629/activity

kevinveenbirkenbach:
image

kevinveenbirkenbach:
https://cloud.infinito.nexus/apps/files/files/270127?dir=/Infinito.Nexus/IT/Architecture/CORE&editing=false&openfile=true

